PR #2, opened 2019-04-28
[WIP] Add preliminary license conversion and tests

Korusuke:
This is by no means complete and would require changes mainly in the list of license and their mapping for which I would like help as I am not sure about all the mappings, as of now for testing I have added a few licenses.

It converts spdx lixenses as per MacPorts naming convention.
Although upt assigns a different name for some licenses, I am not using that and directly using the spdx identifier.


mojca (commented):
Great, thank you very much.

I leave it up to @reneeotten to test and @Steap to provide any feedback about upt integration itself, just two general suggestions:
* I would probably create a json file (or any other simple format, even if just `<spdx identifier><tab><macports identifier>`), so that the same "database" could potentially be used in other parts of MacPorts later on, and to make it much easier to edit and reuse it.
  * I'm perfectly OK with starting with a very small list of licences which you checked (and maybe one of us double-checked), and potentially comment out or otherwise mark or list the ones you didn't yet check (OK, json is admittedly bad in handling comments). Then merge the functional changes with a small correct list, while the rest of the list can be completed later.
* Just for your info for the future: GitHub recently introduced draft pull requests (I didn't know it until another student opened a draft PR). So if you know that the PR is not yet ready to be merged directly, instead of using `[WIP]`, you can open PR as draft and change it to a real one at any later point.

I tried to find the code handling the licences in MacPorts, but I would probably need some more time ... The only thing that the code does at the moment is decide whether we are allowed to mirror the sources and the binaries, and we have some "licence compatibilities table" somewhere in the source.

In the future I would greatly support to adopt the SPDX convention for licence names in all ports (this might require changes in the base), but that's of course outside of the scope of upt. At the time when MacPorts was written SPDX didn't exist yet. Their oldest news post is from 2013. Me and @neverpanic both first heard about it for the first time two years ago at GSOC meeting.

reneeotten:
I agree with @mojca ; we don't want to have the conversion list in the Python file, ```json``` would probably be best.

I think we can start with having only the ones that are converted in that list and leave the "unknown" ones out. You anyway do a ```try.... except``` thing and then set it to "unknown" if the license isn't in the list. The implementation in in ```upt_macports.py``` looks okay; please do make sure though that your Python code follows the style-conventions (i.e., run ```flake8``` or your favorite linter on the code before committing). I will attempt to find some time this week to give it a try!

upt_macports/upt_macports.py
```diff
@@ -28,6 +29,408 @@ def _render_makefile_template(self):
         template = env.get_template(self.template)
         return template.render(pkg=self)
 
+    def spdx_to_macports(self,spdx_license):
```
Steap: If you do not use the "self" parameter, this should be a static method. There should be a whitespace after the comma. If this function is "private", it should start with an underscore, so:

> @staticmethod
> def _spdx_to_macports(self, spdx_license):
> ...

upt_macports/upt_macports.py
```diff
@@ -28,6 +29,408 @@ def _render_makefile_template(self):
...
+                    "psutils": "unknown",
+                    "wxWindows": "unknown",
+                    "xinetd": "unknown",
+                    "xpp": "unknown",
+                    "zlib-acknowledgement": "unknown"
+                }
+        try:
+            macports_license = licenses[spdx_license]
+        except:
+            macports_license = "unknown"
```
Steap: "try: ... except: ..." is frowned upon. You usually want to explicitely list the exceptions you want to catch.

Here, even better than a try/except, you could use dict.get:

> macports_license = licenses.get(spdx_license, 'unknown')

Note that single quotes are preferred in Python.

upt_macports/upt_macports.py
```diff
@@ -28,6 +29,408 @@ def _render_makefile_template(self):
...
+        return macports_license
```
Steap: If you follow the above piece of advice, you could even get rid of the variable:

> return licenses.get(spdx_license, 'unknown')

upt_macports/upt_macports.py
```diff
@@ -28,6 +29,408 @@ def _render_makefile_template(self):
...
+
+    def _licences(self):
+        for i in self.upt_pkg.licenses:
+            i.name = self.spdx_to_macports(i.spdx_identifier)
+        return self.upt_pkg.licenses
```
Steap: When used as a loop index, "i" usually refers to a number (as in "for i in range(42)"). Use a more explicit name, such as "license". It is also a bit weird to modify self.upt_pkg.licenses. This could become:

> return [self.spdx_to_macports(license.spdx_identifier) for license in self.upt_pkg.licenses]

Steap (changes requested):
Thanks for the patch, see my inlined comments. Also, we are going to need tests :) See those in upt-openbsd for instance.

Renee suggests that you run "tox -epy37,flake8". Github allows users to use Travis to run such tests on merge requests. Why don't you submit a patch that adds a Travis configuration file to upt-macports?

Korusuke:
Thanks for the review
I have made all the changes suggested by @reneeotten and @mojca except for deleting the "unknown" ones out which I will do after adding some more to it. (I forgot about it until writing this out)

As suggested by @Steap I have added .travis.yml file and also enabled Travis CI on my fork.

Will add some tests as soon as possible.

upt_macports/upt_macports.py
reneeotten: I agree, and this would have shown up with static code analysis.

upt_macports/upt_macports.py
reneeotten: I have to admit that I've never really looked into this decorators or different method types, so I'll follow your lead here (it's a good opportunity for me to learn along the way as well). 

Please correct me if I am wrong, but if this is a ```staticmethod``` doesn't that imply that it will not use ```self``` and therefore that shouldn't be an argument anymore?

Steap: No you're right :) The whole point was to remove "self", and I failed :) This should be:

> @staticmethod
> def _spdx_to_macports(spdx_license):
> ...

.travis.yml
```diff
@@ -0,0 +1,11 @@
+sudo: false
+dist: xenial
+language: python
+python:
+    - "3.7"
+
+install:
+    - pip install tox-travis
+    - pip install flake8
```
Steap: I think tox would install flake8. Not sure whether tox-travis is needed.

Steap:
Thanks for the updated patch.

 A few things. First, try to only update a patch when it is ready to be reviewed.

Second, try to keep patches as small and "self-contained" as possible. Here, this PR should really be split into three different PR:

1) Fix the pep8 issues
2) Add Travis support
3) Add support for licenses

mojca:
Personally I don't have a problem with a single PR with three "self-contained" commits. In cases where one really need the first patch to have the third commit working, it's more annoying to work with multiple PRs. I would be in favour of multiple commits when the work is completely unrelated and the order of accepting those PRs does not matter at all.

upt_macports/upt_macports.py
```diff
@@ -88,13 +103,15 @@ def _pkgname(self):
...
+        name = name  # for perl changing to lower case may cause error
         return name
 
+
 class MacPortsRubyPackage(MacPortsPackage):
     template = 'ruby.Portfile'
 
-class MacportsBackend(upt.Backend):
+
+class MacPortsBackend(upt.Backend):
```
reneeotten, line 97: if you change the capitalization here (I do agree with that BTW), then I think you'll have to do that as well in in https://github.com/macports/upt-macports/blob/20b55dcc6a06f48da8a22c415bc74315dda5cb2e/setup.cfg#L26

reneeotten (changes requested):
sometimes one PR with a few, small "self-contained" commits might be okay, but if the commits are very different in purpose it's certainly easier to review in separate PRs and/or quickly merge finished commits while still working on the larger issues. So I guess I partially agree with both @Steap and @mojca... What about we finish up this PR with the slightly different topics combined, and in the future err on the side of  having more "one-purpose PRs"?

For example, in the commit "Add MacPorts Backend tests" you change the name of the class in ```upt_macports.py```. That should ideally not be part of the "add tests" commit... and, more importantly, should also be changed in the ```setup.cfg``` file - right?

upt_macports/upt_macports.py
Steap: You're right about that. I missed this by the way :)

This is also why I'd rather see smaller "self-contained" PRs: they are easier to review in github.

Korusuke: So I guess instead of doing different PR from next time, seeing that few of the current commits can be merged and few not.
If everyone agrees then I'll close this PR and open up different PRs for different purposes namely:
- Add Travis support
- Add license conversion
- Add tests

mojca: My 2 cents. I agree that it generally makes more sense to do small and self-contained pull requests on independent code, so if one thing is ok, it can be merged immediately, and the other one can be fixed / discussed later.

But on the other hand you started this PR to do one feature, and then it turned out that some more stuff needs to be done to consider this feature complete. If you write ten lines per day this can be done, but if one produces hundreds of lines of code per day with dependencies (feature A needs to be merged before feature B), it becomes a mess to deal with a zillion pull requests, and I see no reason to be overly draconic about what goes in what pull request, and at the end produce 10 times less code than if allowed to work efficiently.

From my point of view it makes more sense to just clean the commits and not open new pull requests for its own sake.

reneeotten: @mojca @Korusuke yes, that's what I said - finish up this PR with separating commits into logical changes for now and move on ;) Let's just keep this discussion in mind for the future and be flexible if a single-purpose PR evolves into something more!

Korusuke: Okeyy then I will just separate all out into different commits and push the changes soon.

neverpanic:
JFYI, the place to change so that MacPorts' infrastructure can correctly deal with SPDX license identifiers is https://github.com/macports/macports-infrastructure/blob/master/jobs/port_binary_distributable.tcl.

upt_macports/spdx2macports.json
```diff
@@ -0,0 +1,42 @@
+{
```
reneeotten, line 1: this list doesn't need to be complete, but now there are a few very common licenses missing. I think we should add for sure:
- MIT License
- the "GNU Library General Public License" family

upt_macports/upt_macports.py
```diff
@@ -29,6 +31,16 @@ def _render_makefile_template(self):
         template = env.get_template(self.template)
         return template.render(pkg=self)
 
+    @property
+    def licenses(self):
+        path = 'spdx2macports.json'
+        filepath = pkg_resources.resource_filename(__name__, path)
+        with open(filepath) as f:
+            spdx2maports = json.loads(f.read())
```
reneeotten: this variable should be called ```spdx2macports``` here and in the next line

upt_macports/upt_macports.py
```diff
@@ -29,6 +31,16 @@ def _render_makefile_template(self):
...
+
+        return ' '.join([spdx2maports.get(license.spdx_identifier, 'unknown')
+                        for license in self.upt_pkg.licenses])
+
```
reneeotten, line 38: you didn't update the ```templates/base.Portfile``` so I it's not actually doing anything with the converted licenses as far as I can tell. For example, if I install this PR in a virtualenv and do ```upt package -f pypi -b macports upt``` it still shows me ```3-clause BSD License``` as the license. Please check.

upt_macports/upt_macports.py
```diff
@@ -29,6 +31,16 @@ def _render_makefile_template(self):
         template = env.get_template(self.template)
         return template.render(pkg=self)
 
+    @property
+    def licenses(self):
+        path = 'spdx2macports.json'
```
reneeotten: the name ```path``` is a bit strange here, as it's actually the filename - or?

reneeotten (changes requested):
hi Karan, 

the separation into individual commits looks good to me. However, I don't think that the actual license conversion works as intended, can you please check that?

upt_macports/upt_macports.py
```diff
@@ -2,15 +2,16 @@
...
         self.logger = logging.getLogger('upt')
 
     def create_package(self, upt_pkg, output):
         self.upt_pkg = upt_pkg
-        #self._create_output_directories(upt_pkg, output)
-        #self._create_makefile()
-        #self._create_pkg_descr()
+        # self._create_output_directories(upt_pkg, output)
+        # self._create_makefile()
+        # self._create_pkg_descr()
```
Steap: If we are going to clean up, remove these lines.

upt_macports/upt_macports.py
```diff
@@ -88,13 +91,15 @@ def _pkgname(self):
 
     @staticmethod
     def _normalized_macports_name(name):
-        name = name # for perl changing to lower case may cause error
+        name = name  # for perl changing to lower case may cause error
         return name
```
Steap, line 90: This should just be "return name". "name = name" is a no-op.

upt_macports/upt_macports.py
```diff
@@ -29,6 +31,16 @@ def _render_makefile_template(self):
         template = env.get_template(self.template)
         return template.render(pkg=self)
 
+    @property
+    def licenses(self):
+        path = 'spdx2macports.json'
+        filepath = pkg_resources.resource_filename(__name__, path)
```
Steap: When you run "python setup.py install", is spdx2macports.json installed? Don't you need to mess with MANIFEST.in or something?

Steap (changes requested):
The tests should be merged in the same commit as the feature. Please squash "Add support for license conversion" and "Add License tests".

upt_macports/upt_macports.py
Korusuke: It's actually the relative path but since both files are in the same place, it finally is just filename.
Did not want to name it filename as that would be a bit misleading.
How about `relpath` or `relativepath` ?

upt_macports/upt_macports.py
Korusuke: I have applied these changes and other suggested below, sorry for the silly mistakes :)

upt_macports/upt_macports.py
reneeotten: I am not sure that it would be misleading, isn't this always the filename and does ```pkg_resources.resource_filename``` get the (relative) path to this filename? Or am I missing something here (@Steap)?

upt_macports/spdx2macports.json
reneeotten: thanks for adding ```MIT```. For the others I was talking about all the SPDX identifiers with ```GPL-*``` and ```LGPL-*```, see [here](https://spdx.org/licenses/). I think after adding those we have (at least for now) a complete enough list of licenses.

When you add those licenses, please also rename the last commit to something along the lines ```Add support for license conversion and include tests". You did combine the commits as @Steap suggested, but didn't update the commit message.

reneeotten:
@Korusuke @Steap this should be almost ready to merge I think!

Steap:
> @Korusuke @Steap this should be almost ready to merge I think!

I think the first two commits could be pushed, yes. Is there any way to do that? It would be nice to see what Travis has to say about the other two.

Korusuke:
Travis is already active on my fork if you want to have a look.

Korusuke:
@Steap @reneeotten Is there anything else remaining or is this PR ready to be merged?

Also once this is merged how should we proceed forward... as per the proposal this time period is for discussion on port updates, livecheck and also npm packages.

reneeotten:
> @Steap @reneeotten Is there anything else remaining or is this PR ready to be merged?
> 
> Also once this is merged how should we proceed forward... as per the proposal this time period is for discussion on port updates, livecheck and also npm packages.

As far as I am concerned this can be merged and I will do so today/tomorrow unless @Steap has concerns left with the content. Of course the license conversion will need testing "in the wild" and might need some fine-tuning based on that, but we'll never know before actually using it. 

Regarding on how to move forward... I'll look this weekend at the proposal timeline and topics you mentioned, gather my thoughts and will get back to you. Also, I do have a few "feature request"-like things on a piece of paper somewhere, which I will probably add as "Issues" here on this repository just so we don't forget; some of them can probably addressed rather easily anyway.

Steap:
The PR looks good to me, I'll merge it.

reneeotten:
@Korusuke thanks! Please do open a PR in macports-ports to update the py-upt-macports port (and also for the py-upt port to version 0.7)

Korusuke:
> @Korusuke thanks! Please do open a PR in macports-ports to update the py-upt-macports port (and also for the py-upt port to version 0.7)

Done

Issue #3, opened 2019-05-15
set keyword depending on archive compression type


reneeotten:
If the archive is not a ```.tar.gz``` file, but uses a different compression algorithm one should set the proper keyword in the Portfile (see [here](https://guide.macports.org/#reference.phases.extract) for the options).

It appears that this information is available from ```upt```, probably from ```upt.upt_pkg.archives```. It can be extracted from the ```_filename``` attribute or perhaps better ```archive_type```, even though I am not sure that this is set correctly (doesn't appear to be the case for ```py-algopy```). Examples of ports where this is the case are ```py-algopy``` for the ```pypi``` frontend and ```p5-term-gnuplot``` for the ```cpan``` frontend.

Issue #4, opened 2019-05-15
layout base.Portfile


reneeotten:
A few suggestions for the ```base.Portfile``` template:

1. adding a comment along the lines of "Portfile generated/maintained using the Universal Packaging Tool (upt)" would be useful to indicate which ports are using ```upt```. (I would add it just before the ```name``` field).
2. rephrase the comment about ```supported_archs``` for more clarity. How about something like "# uncomment the following line if no architecture-dependent files are installed, otherwise remove"
3. remove lines 16 (```print(self._depends)```) and 25 (```# env.filters['reqformat'] = self.jinja2_reqformat```) as they seem unnecessary 
4. add ```categories` field (probably commented out since it's difficult to fill in sensible values automatically)
5. add "default" example for "maintainers"; ```maintainers: {example.org:julesverne @GitHub-handle} openmaintainer```
6. follow the order of fields from [the guide](https://guide.macports.org/#development.creating-portfile) a bit more closely. Admittedly, some of this is personal preference so it doesn't _need_ to be changed, but please have a look. I would recommend the following order:

```
# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4

PortSystem          
PortGroup           

name                
version             
revision            

# categories-append
platforms           
# uncomment the following line if no architecture-dependent files are installed, otherwise remove
# supported_archs     noarch
license            
maintainers: {example.org:julesverne @GitHub-handle} openmaintainer        

description         
long_description    

homepage            
master_sites        
distname            
# if needed, add keyword depending on compression type of archive
checksums           
```

PR #7, opened 2019-05-17
Add archive type keyword

Korusuke:
This addresses Issue #3 

Since upt is setting archive_type to targz for all sdist, so instead of using that extracting from URL seems better.
@Steap Should this feature be directly implemented in upt itself or is this fine?

Once this is merged I will push the changes for #4

### DIFF
--- a/upt_macports/templates/base.Portfile
+++ b/upt_macports/templates/base.Portfile
@@ -16,7 +16,13 @@ platforms           darwin
 # supported_archs     noarch
 
 {% block dist_info %}
-{% endblock %}
+{% endblock -%}
+
+{% if pkg.archive_type == 'unknown' -%}
+# Detection of archive type failed
+{% elif pkg.archive_type != 'gz' %}
+{{ "use_%-15s yes"|format(pkg.archive_type) }}
+{% endif %}
 
 licenses            {{ pkg.licenses }}
 
--- a/upt_macports/tests/test_macports_package.py
+++ b/upt_macports/tests/test_macports_package.py
@@ -32,5 +32,26 @@ class TestMacPortsPackageLicenses(unittest.TestCase):
         self.assertEqual(self.package.licenses, expected)
 
 
+class TestMacPortsPackageArchiveType(unittest.TestCase):
+    def setUp(self):
+        self.package = MacPortsPackage()
+        self.package.upt_pkg = upt.Package('foo', '42')
+
+    def test_no_archive(self):
+        self.package.upt_pkg.archives = []
+        expected = 'unknown'
+        self.assertEqual(self.package.archive_type, expected)
+
+    def test_known_archive(self):
+        self.package.upt_pkg.archives = [upt.Archive("url.co/dir/file.tar.gz")]
+        expected = 'gz'
+        self.assertEqual(self.package.archive_type, expected)
+
+    def test_unknown_archive(self):
+        self.package.upt_pkg.archives = [upt.Archive("url.co/dir/file.rar")]
+        expected = 'unknown'
+        self.assertEqual(self.package.archive_type, expected)
+
+
 if __name__ == '__main__':
     unittest.main()
--- a/upt_macports/upt_macports.py
+++ b/upt_macports/upt_macports.py
@@ -51,6 +51,26 @@ class MacPortsPackage(object):
     def test_depends(self):
         return self._depends('test')
 
+    @property
+    def archive_type(self):
+        archive_keyword = {
+            'gz': 'gz',
+            '7z': '7z',
+            'bz2': 'bzip2',
+            'lzma': 'lzma',
+            'tar': 'tar',
+            'zip': 'zip',
+            'xz': 'xz'
+        }
+        try:
+            archive_name = self.upt_pkg.get_archive().filename
+            archive_type = archive_name.split('.')[-1]
+            return archive_keyword.get(archive_type, 'unknown')
+
+        except upt.ArchiveUnavailable:
+            self.logger.error('Could not determine the type of the source archive') # noqa
+            return 'unknown'
+
 
 class MacPortsPythonPackage(MacPortsPackage):
     template = 'python.Portfile'


reneeotten:
> Since upt is setting archive_type to targz for all sdist, so instead of using that extracting from URL seems better. @Steap Should this feature be directly implemented in upt itself or is this fine?

That's what appeared to happen from a quick look at the ```upt``` code. I would think the ```archive_type``` could/should be set correctly in ```upt```, but let's see what @Steap thinks about this. Otherwise getting the information from the URL and doing this in the MacPorts backend would work as well.

Why the change in line 19 to ```{% endblock -%}``` ; what does the ```-``` sign do?

Korusuke:
The `-` sign is just for deleting newline after that block as if it's not there there would be an extra line after `distname` in python ports and two blank lines after `supported_archs` in perl ports.

reneeotten:
> The `-` sign is just for deleting newline after that block as if it's not there there would be an extra line after `distname` in python ports and two blank lines after `supported_archs` in perl ports.

Aha, okay... so that's how to avoid empty lines, thanks for the explanation ;) Let's see if the idea is to get the ```archive_type``` set in ```upt``` or if we should do this in the backend.

Steap:
Hey! Thanks for the PR.

So, in upt, the archive_type is set to SOURCE_TARGZ even when upstream uses a ZIP file, for instance. I'll probably rename it to "SOURCE" so that it is less confusing. I could encode the filetype, but I think backends would still need to have a long "switch/case" to handle all possible cases, so that may not help that much.

Your patch looks good, but it could crash in some cases. upt_pkg.archives might be an empty list, so you'd get an IndexError when accessing the first element. You should probably move this code to upt_macports.py and have lighter code in the template. This would also make it easier to write tests. You may also want to use "upt_pkg.get_archive()" which will return the source archive, rather than archives[0].

Korusuke:
Initially, I had written this code in `upt_macports.py` and was passing just the compression type such as `zip`, `xz`, etc. and then in the template was just appending the type to `use` and writing `yes` there.
But the problem with that was
If the type is zip we need:
<pre>
usezip          yes
</pre>
and if xz then:
<pre>
usexz           yes
</pre>
but instead what i was getting was:
<pre>
usexz            yes
</pre>
as the spaces between usezip, usexz and yes is different that was causing issues, I will look into if there's any solution this.
If I don't find anything then the py script itself can return the whole thing like `usexz           yes` 


Also about the crash due to archive[0], It will cause the same issue in checksums, right?
`pkg.upt_pkg.archives[0].sha256`

reneeotten:
@Korusuke you can the [```.format()``` method](https://docs.python.org/3/library/string.html#format-string-syntax) and specify how to align things, so that the ```yes``` is always aligned correctly.

reneeotten:
@Korusuke from a quick check it appears to be working as intended. However, I would encourage you to take another look at the ```archive_type``` method you added and see if that could be written in a more compact/Pythonic way. 

To get you started, I would think about it like this:
- let's get the extension of the file (i.e., the part after the last ```.```)
- then there are three scenarios: (i) the extension is in a group/list that can be used directly in the ```use_extension```, (ii) it's ```bz2``` - in that case it should be ```bzip2``` instead of ```bz2```, (iii) it's not part of the pre-defined MacPorts archive types, then return "unknown" or something like that.

In the template there are then two scenarios: (i) the archive_type/extension is unknown, (ii) it is one of the pre-defined ones add ```use_extension``` to the template unless it's ```.gz```.

upt_macports/tests/test_macports_package.py
```diff
@@ -32,5 +32,21 @@ def test_multiple_license(self):
         self.assertEqual(self.package.licenses, expected)
 
 
+class TestMacPortsPackageArchiveType(unittest.TestCase):
+    def setUp(self):
+        self.package = MacPortsPackage()
+        self.package.upt_pkg = upt.Package('foo', '42')
+
```
Steap, line 39: Add a test with:

>  self.package.upt_pkg.archives = []

See what happens.

upt_macports/upt_macports.py
```diff
@@ -51,6 +51,26 @@ def run_depends(self):
...
+            return 'lzma'
+        elif archive_url.endswith('.tar'):
+            return 'tar'
+        elif archive_url.endswith('.zip'):
+            return 'zip'
+        elif archive_url.endswith('.tar.xz'):
+            return 'xz'
+        else:
+            return 'undefined'
+
```
Steap, line 73: This is fine, but maybe we could do something like;:

```
mappings = {
    '.tar.gz': 'targz',
    '.7z': '7z',
    ...
}
...
for extension, fmt in mappings.items():
    if archive_url.endswith(extension):
        return fmt
return 'undefined'
```

What do you think? There might be better variable names :)

Steap (changes requested):

Steap:
> * let's get the extension of the file (i.e., the part after the last `.`)

In the case of "foo-13.37.tar.bz2", the extension is actually "tar.bz2", so it is not always the part after the last '.', or am I misunderstanding something?

upt_macports/templates/base.Portfile
```diff
@@ -16,7 +16,14 @@ platforms           darwin
 # supported_archs     noarch
 
 {% block dist_info %}
-{% endblock %}
+{% endblock -%}
+
+{% if pkg.archive_type == 'targz' -%}
```
Steap: This empty if is weird, maybe you could...

upt_macports/templates/base.Portfile
```diff
@@ -16,7 +16,14 @@ platforms           darwin
...
+{% elif pkg.archive_type == 'undefined' %}
+# Archive type could not be detected by upt
+{% else %}
+{{ "use_%-15s yes"|format(pkg.archive_type) }}
```
Steap, line 24: ... use "{{ if pkg.archive_type != 'targz' }}" here. What do you think?

reneeotten:
> > * let's get the extension of the file (i.e., the part after the last `.`)
> 
> In the case of "foo-13.37.tar.bz2", the extension is actually "tar.bz2", so it is not always the part after the last '.', or am I misunderstanding something?

You're certainly right about that, but please take a look in the [MacPorts guide](https://guide.macports.org/#reference.phases.extract) and how the ```use_???``` are defined. You'll see that ```use_bzip2``` considers ```tar.bz2``` as the ```extract.suffix``` and ```use_xz``` corresponds to ```tar.xz```. So for all practical purposes I think that the thing we care about here is "the part after the last ```.```); of course we don't have to call it the file "extension" ;)

upt_macports/upt_macports.py
Korusuke: If using just the last `.` is fine then maybe we can do just this :

```
def archive_type(self):
        archive_keyword = {
            'gz': 'targz',
            '7z': '7z',
            'bz2': 'bzip2',
            'lzma': 'lzma',
            'tar': 'tar',
            'zip': 'zip',
            'xz': 'xz'
        }
        archive_url = self.upt_pkg.get_archive().url
        archive_type = archive_url.split('.')[-1]

        return archive_keyword.get(archive_type, 'unknown')
```

upt_macports/tests/test_macports_package.py
Korusuke: It raises an ArchiveUnavailable error as expected, I have added a test for the same.

```
def test_no_archive(self):
        self.package.upt_pkg.archives = []
        with self.assertRaises(upt.ArchiveUnavailable):
            self.package.archive_type
```

Steap: But should "self.package.archive_type" raise an exception? Shouldn't it rather handle it and return 'unknown'?

upt_macports/upt_macports.py
Steap: Yeah, obviously using dict.get() is better :)

Steap:
> > > * let's get the extension of the file (i.e., the part after the last `.`)
> > 
> > 
> > In the case of "foo-13.37.tar.bz2", the extension is actually "tar.bz2", so it is not always the part after the last '.', or am I misunderstanding something?
> 
> You're certainly right about that, but please take a look in the [MacPorts guide](https://guide.macports.org/#reference.phases.extract) and how the `use_???` are defined. You'll see that `use_bzip2` considers `tar.bz2` as the `extract.suffix` and `use_xz` corresponds to `tar.xz`. So for all practical purposes I think that the thing we care about here is "the part after the last `.`); of course we don't have to call it the file "extension" ;)

OK, makes sense.

upt_macports/tests/test_macports_package.py
Korusuke: If we handle the exception and generate the portfile then many details will be missing as checksums will be missing/wrong too.
So should we really generate the portfile when the main thing is missing?
I agree that at least a few details will be there and that can be used as an outline by the maintainer.

Steap: The checksums would be missing, not wrong.

We should probably "do our best". The goal of upt is to create roughly 90% of the package. If for some reason we can only create 50%, well, the maintainer still has something to work with, and saved a few minutes of their life.

We should probably log this as an error, though.

upt_macports/upt_macports.py
reneeotten: Agreed! For the sake of consistency I would just use ```'gz': 'gz'``` as this is no different than for example the situation for ```xz```.

upt_macports/tests/test_macports_package.py
reneeotten: > We should probably log this as an error, though.
Yes, that would be good indeed (there are probably other instances to log as error/warning; for example when the license cannot be translated to MacPorts). Should that be part of this PR or should we look at the logging by ```upt_macports``` separately?

Korusuke: Ok so if the archive is not found then at this point should we log 'Archive not found on upstream' or should we say 'Archive detection failed as no archive found' as the latter is appropriate as per the place where it is being logged but even the former is not wrong.

And also as of now, I am logging it like this 
```
try:
            archive_url = self.upt_pkg.get_archive().url
            archive_type = archive_url.split('.')[-1]
            return archive_keyword.get(archive_type, 'unknown')
except upt.ArchiveUnavailable:
            self.logger.error(f'Archive detection failed as no archive found')
            return 'unknown'
```
and the test has been changed accordingly.
```
self.package.upt_pkg.archives = []
expected = 'unknown'
self.assertEqual(self.package.archive_type, expected)
```

Is it fine now? :)

Also as of the other logging goes I guess from now on for every feature wherever required we will add the logging and for the existing code, we can add the logging in another PR.

reneeotten: hi @Korusuke, ehm I don't really see the change you mention above in the PR... but perhaps you were waiting for a response before adding it?

I would like to avoid two ```return``` statements, so what about:
```
try:
    archive_url = self.upt_pkg.get_archive().url
    archive_type = archive_url.split('.')[-1]
except upt.ArchiveUnavailable:
    self.logger.error(f'Detection of archive type failed: no archive found')
    archive_type = None
finally:
    return archive_keyword.get(archive_type, 'unknown')
```

Additionally, as I commented earlier I would not treat ```gz``` differently in the ```archive_keyword``` dictionary than ```xz```; just return ```gz``` there as well.

reneeotten (changes requested):
This should be the last round of comments - I'd suggest to wait for @Steap his reaction so everything can incorporated at once and let's get this merged!

@Korusuke thanks for the quick turnaround of changes and I'm sorry for suggesting more changes... There are always multiple ways of coding things and even though it takes a little longer before merging, it never hurts to discuss a bit and finish up with the "best" solution!

upt_macports/tests/test_macports_package.py
Steap: So we could use "self.upt_pkg.get_archive().filename" rather than "self.upt_pkg.get_archive().url", but both work :)

I like the simple try/except better than the try/except/finally with a single return statement, but I'm ok with both as long as we have good tests.

upt_macports/upt_macports.py
Steap: I'm confused, don't we have a special case for "tar.gz" files?

Steap:
@Korusuke I agree with most of Renee's comments, so feel free to update the PR.

Steap:
Oh btw, can't we add "Fixes: #3" in the commit log so that the bug is closed automatically once this is merged?

upt_macports/tests/test_macports_package.py
reneeotten: fine with me, leave the ```try....except``` as is, but please do change the logging message. Using ```self.upt_pkg.get_archive().filename``` would indeed make sense.

upt_macports/upt_macports.py
reneeotten: Yes, we do.

What I mean is that what is returned shouldn't be ```targz``` in my opinion (if we follow that logic ```xz``` should return ```tarxz```. So what I am proposing is -for consistency - to change ```'gz': 'targz',``` to ```'gz': 'gz',``` and in the template use ```{% elif pkg.archive_type != 'gz' %}```

upt_macports/tests/test_macports_package.py
Korusuke: Yeah using `self.upt_pkg.get_archive().filename` make more sense, I had missed that it exists as my local copy of upt was old and did not have the filename, filepath properties :)

upt_macports/upt_macports.py
```diff
@@ -51,6 +51,27 @@ def run_depends(self):
...
+            'xz': 'xz'
+        }
+        try:
+            archive_name = self.upt_pkg.get_archive().filename
+            archive_type = archive_name.split('.')[-1]
+            return archive_keyword.get(archive_type, 'unknown')
+
+        except upt.ArchiveUnavailable:
+            self.logger.error(f'Detection of archive type failed: no archive \
+found')
```
reneeotten: I am not a fan of this indentation, but I know you did that to get the text continued correctly. I would not be against just keeping this on one line and violating the "80-length" rule a bit, but perhaps the tests would complain about that....  Another option would be:
```
self.logger.error(f'Detection of archive type failed: no archive ' \
                   'found')
```

besides that (even with this) I would be happy to merge!

mojca:
In the test you can explicitly say that it's ok to violate the line length.

upt_macports/upt_macports.py
Korusuke: Ok so I will keep it normal (violate 80-length) as that looks better and then maybe in other PR or if you are fine then in this same PR change the test to allow that rule violation as @mojca said.

upt_macports/upt_macports.py
```diff
@@ -51,6 +51,26 @@ def run_depends(self):
...
+            'zip': 'zip',
+            'xz': 'xz'
+        }
+        try:
+            archive_name = self.upt_pkg.get_archive().filename
+            archive_type = archive_name.split('.')[-1]
+            return archive_keyword.get(archive_type, 'unknown')
+
+        except upt.ArchiveUnavailable:
+            self.logger.error(f'Detection of archive type failed: no archive found')
```
Steap: Do not use an f-string when it is not needed. I also think this should rather read something like "Could not determine the type of the source archive". Writing "no archive found" makes me think upt-macports was looking for the archive, while it was upt-pypi which failed to find one.

Also, if the line is too long, you can add "# noqa" at the end to disable the linter for this particular line.

Korusuke: > Also, if the line is too long, you can add "# noqa" at the end to disable the linter for this particular line.

Oo I had no idea about this, I thought global test settings bust be changed.

I have done the new changes as suggested.

reneeotten (approved):

reneeotten:
Great - a concise function to do the work, tests added to make sure everything works and an updated template! What else can one request - in it goes ;)

Issue #8, opened 2019-05-18
Tests: use coverage


Steap:
It would be nice to use "coverage" in the tests. Upt does it (see https://framagit.org/upt/upt/commit/3dfc1b4c7b3a163212754538b973503263759b6c) and it helps detect which parts of the code are never executed when running the tests.

reneeotten:
Sure, that never hurts! 

In the same PR to add this, let's also fix the configuration for the Azure pipelines that Rainer enabled. Configuration file is ```azure-pipelines.yml``` and the documentation is [here](https://docs.microsoft.com/en-us/azure/devops/pipelines/languages/python?view=azure-devops).

PR #9, opened 2019-05-18
Add coverage and update azure config

Korusuke:
This addresses Issue #8 

I have added coverage but as of now, it will fail as the pass criteria is 90% and till now the coverage is at 65%.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,6 @@
+[run]
+branch = True
+
+[report]
+omit = upt_macports/tests/*
+fail_under = 60
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -6,18 +6,23 @@
 trigger:
 - master
 
-pool:
-  vmImage: 'ubuntu-latest'
 strategy:
   matrix:
-    Python27:
-      python.version: '2.7'
-    Python35:
-      python.version: '3.5'
-    Python36:
+    py36_linux:
+      imageName: 'ubuntu-16.04'
       python.version: '3.6'
-    Python37:
+    py37_linux:
+      imageName: 'ubuntu-16.04'
       python.version: '3.7'
+    py36_macos:
+      imageName: 'macos-10.13'
+      python.version: '3.6'
+    py37_macos:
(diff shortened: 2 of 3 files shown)


Steap:
I would like this PR to be split into two different commits.

You can put the threshold at 60% in .coveragerc, and add more tests later.

reneeotten:
I agree with @Steap, the Azure part should be in a different commit. 

You should for sure take a look at the Python versions that you have selected in the ``` azure-pipelines.yml```, ```upt``` will only work with Python >= 3.6. Additionally, I don't see you installing ```upt```, the frontends and ```upt-macports``` backend, so it's not really surprising that the tests fail ;) [edit: sorry, ignore that last comment ```tox``` should take care of this]

The idea of using Azure for the ```macports-port``` repository was that it allows for longer builds (Travis has a maximum of 50 minutes), which is important for ports that can take a long time to build. However, the tests here are finished very quickly so we'll never run into timeout issues with Travis; thinking about this a bit more, perhaps there is no reason to actually run the tests on Azure as well?

mojca:
A few notes about Azure. For the ports it was absolutely critical to enable it as Travis is basically useless most of the time. But at the same time Travis fired most of their developers, so in my opinion it's only a question of time when they shut down for free projects. It absolutely doesn't hurt to have both enabled.

I certainly consider it optional to have two different services enabled, but now that the code is there, there should be no reason not to use it.

Also, there's nothing wrong with running the tests on Ubuntu (that is: both GNU Linux and macOS). Optional of course, but if it's just a matter of two lines in configuration, it cannot hurt to test on both.

Korusuke:
As suggested I have done the following changes:
- Split into two commits
- Change threshold to 60
- Add macOS and ubuntu both for py36 and py37

Now the checks are passing :) so I guess this commit may be good to merge.

reneeotten (approved):

reneeotten:
@Steap - are you okay with merging this? If I don't hear otherwise, I'll merge this tomorrow morning.

Steap:
I trust with you the Azure stuff :)

Issue #5, opened 2019-05-15
whitespace/layout changes to python.Portfile template


reneeotten:
Ideally, the layout/whitespaces within the ```if {${name} ne ${subport}} { }``` block should be the same irrespective of which/how many dependency types are present.

Try packaging ```setuptools```, ```upt``` and ```pytest``` as examples of the different scenarios and make sure the layout is the same as below (i.e., no newline at the beginning of the if-block and one newline between the different dependency types, and the livecheck):
```
if {${name} ne ${subport}} {
    depends_build-append \
                    port:py${python.version}-setuptools

    depends_lib-append \
                    port:py${python.version}-???

    depends_test-append \
                    port:py${python.version}-???

   livecheck.type  none
```

Steap:
Is there a linter that could help us check that the spacing is right?

reneeotten:
> Is there a linter that could help us check that the spacing is right?

no, there is not. ```port lint --nitpick <portname>``` will detect several issues, but regardless of the whitespace the syntax of the Portfile is correct. It's just that for most Portfiles this whitespacing is used and (in my opinion) it also adds to the readability so I would like to have ```upt``` generate it like this.

Steap:
Fixed in  156b5ed.

PR #10, opened 2019-05-18
Update Layout for python portfile

Korusuke:
Issue #5 

Two blank lines feel weird but that is what works :)

### DIFF
--- a/upt_macports/templates/python.Portfile
+++ b/upt_macports/templates/python.Portfile
@@ -19,24 +19,28 @@ distname            ${python.rootname}-${version}
 python.versions     37
 
 if {${name} ne ${subport}} {
-
     depends_build-append \
                     port:py${python.version}-setuptools
-    {% if pkg.upt_pkg.requirements.run|length > 0 %}
+    {%- if pkg.upt_pkg.requirements.run|length > 0 %}
+
+
     depends_lib-append \
                     {% for i in pkg.upt_pkg.requirements.run|sort(attribute='name') %}
                     port:py${python.version}-{{i.name}}{% if not loop.last %} \
                     {% endif %}
                     {% endfor %}
     {% endif %}
+    {%- if pkg.upt_pkg.requirements.test|length > 0 %}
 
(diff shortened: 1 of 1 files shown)


Korusuke:
This is the output after current changes:

setuptools :
```
if {${name} ne ${subport}} {
    depends_build-append \
                    port:py${python.version}-setuptools

    livecheck.type  none
}
```

upt :
```
if {${name} ne ${subport}} {
    depends_build-append \
                    port:py${python.version}-setuptools

    depends_lib-append \
                    port:py${python.version}-spdx-lookup

    livecheck.type  none
}
```

pytest :
```
if {${name} ne ${subport}} {
    depends_build-append \
                    port:py${python.version}-setuptools

    depends_lib-append \
                    port:py${python.version}-atomicwrites \
                    ...
                    port:py${python.version}-wcwidth

    depends_test-append \
                    port:py${python.version}-argcomplete \
                    ...
                    port:py${python.version}-requests

    livecheck.type  none
}
```

reneeotten:
> Two blank lines feel weird but that is what works :)

It indeed works, but I agree with you that it looks a little weird in the template. I would need to look a bit more at the ```ninja``` syntax if there is some other way that will give the same output - it not, I guess this would be okay.

reneeotten:
okay, I spend an hour or so looking at the documentation and doing some trial-and-error attempts to get the newlines where we want them..... I was hoping for a more of "what-you-see-is-what-you-get" approach by putting the empty lines where we actually want them, but without much/any success ;) 
There are settings in ```upt_macports.py``` that probably determine some of this stuff for the ```ninja2``` templates, but I haven't figured it out yet.

Since your solution works, despite the slight weirdness of the two empty lines in the template, I am inclined to merge this as "it works and probably nobody is going to change this anymore". Unless @Steap has some insight and we can come up quickly with something else; otherwise let's move on and we can look at this again if we're bored!

Steap:
@reneeotten Yeah, there are also a few variables (such as keep_trailing_newline) that are interesting when creating the Jinja2 environment, but I can never understand/remember how everything works.

Let's merge this :)

PR #13, opened 2019-05-23
Update base layout and remove unnecessary code

### LINKED ISSUE #4: layout base.Portfile

Korusuke:
Fixes #4

### DIFF
--- a/upt_macports/templates/base.Portfile
+++ b/upt_macports/templates/base.Portfile
@@ -4,16 +4,22 @@ PortSystem          1.0
 {% block portgroup %}
 {% endblock %}
 
+# Portfile generated/maintained using the Universal Packaging Tool (upt)
 {% block nameversion %}
 {% endblock %}
 
-maintainers         nomaintainer
+# Please adapt the lines with a "TODO" statement and check the guide for more information:
+# https://guide.macports.org/#reference.keywords
+
+# categories-append # TODO
+platforms           darwin
+# TODO: uncomment the following line if no architecture-dependent files are installed, otherwise remove
+# supported_archs     noarch
+license             {{ pkg.licenses }}
+# maintainers         {example.org:julesverne @GitHub-handle} openmaintainer # TODO
+
 description         {{ pkg.upt_pkg.summary }}
 long_description    ${description}
-
-platforms           darwin
-# uncomment and edit the following line depending on whether the port installs architecture-dependent files
-# supported_archs     noarch
 
 {% block dist_info %}
 {% endblock -%}
@@ -23,8 +29,6 @@ platforms           darwin
 {% elif pkg.archive_type != 'gz' %}
 {{ "use_%-15s yes"|format(pkg.archive_type) }}
 {% endif %}
-
-license             {{ pkg.licenses }}
 
 checksums           sha256  {{ pkg.upt_pkg.archives[0].sha256 }} \
                     rmd160  {{ pkg.upt_pkg.archives[0].rmd160 }} \
--- a/upt_macports/upt_macports.py
+++ b/upt_macports/upt_macports.py
@@ -13,7 +13,6 @@ class MacPortsPackage(object):
         self.upt_pkg = upt_pkg
         self.logger.info(f'Hello, creating the package')
         print(self._render_makefile_template())
-        print(self._depends)
 
     def _render_makefile_template(self):
         env = jinja2.Environment(
@@ -22,7 +21,6 @@ class MacPortsPackage(object):
             lstrip_blocks=True,
             keep_trailing_newline=True,
         )
-        # env.filters['reqformat'] = self.jinja2_reqformat
         template = env.get_template(self.template)
         return template.render(pkg=self)
 


Steap:
It is kind of weird to have a commit titled "Merge branch 'master' into base_port". What happened?

You could add "Fixes #4" at the end of the commit log. I think it should close the corresponding issue automagically once we merge the PR.

reneeotten:
@Korusuke we don't want to have "Merge" commits in the git history.

You probably did ```git pull```; instead you should use ```git pull --rebase``` which will add your new commits on top of the ```master``` branch. Please see the documentation: https://trac.macports.org/wiki/WorkingWithGit#updating. 

Try to recover from this first (a good way of working on your git-vodoo a bit more) and then we can review the actual content. Do make a backup first (```git checkout base_port ; git checkout -b backup```, go back to your branch ```git checkout base_port``` and try to fix it).

Korusuke:
A merge conflict was showing up due to the newly merged #12 so I just tried solving it using the git GUI which now I guess I shouldn't have done, will fix it in few minutes.

reneeotten:
> A merge conflict was showing up due to the newly merged #12 so I just tried solving it using the git GUI which now I guess I shouldn't have done, will fix it in few minutes.

something is not right yet as you have duplicate fields in the template now....

upt_macports/templates/base.Portfile
```diff
@@ -4,16 +4,20 @@ PortSystem          1.0
...
+# Portfile generated/maintained using the Universal Packaging Tool (upt)
+
 {% block nameversion %}
 {% endblock %}
 
-maintainers         nomaintainer
-description         {{ pkg.upt_pkg.summary }}
-long_description    ${description}
-
+# categories-append
```
Steap: You could add "XXX" or "TODO" at the end of this line, so that it is clearly visible to the maintainer when they open the file. @reneeotten what do you think?

reneeotten: I agree - we should signal in a clear way that a maintainer is supposed to do something about it. I think we should add one link to the MacPorts guide on top of that block in the Portfile, comment out lines where action is required, and add TODO at the end of the line. I have made some suggestions in the the base template.

upt_macports/templates/base.Portfile
```diff
@@ -4,16 +4,20 @@ PortSystem          1.0
 {% block portgroup %}
 {% endblock %}
 
+# Portfile generated/maintained using the Universal Packaging Tool (upt)
+
```
reneeotten: let's not add an new line here

upt_macports/templates/base.Portfile
```diff
@@ -4,16 +4,20 @@ PortSystem          1.0
...
 {% block nameversion %}
 {% endblock %}
 
-maintainers         nomaintainer
-description         {{ pkg.upt_pkg.summary }}
-long_description    ${description}
-
```
reneeotten: add one line here to indicate where a maintainer can bind more information: 
```
# Please adapt the lines with a "TODO" statement and check the guide for more information:
# https://guide.macports.org/#reference.keywords 
```

upt_macports/templates/base.Portfile
```diff
@@ -4,16 +4,20 @@ PortSystem          1.0
...
+# categories-append
 platforms           darwin
-# uncomment and edit the following line depending on whether the port installs architecture-dependent files
+# uncomment the following line if no architecture-dependent files are installed, otherwise remove
```
reneeotten: add "TODO", like: ```# TODO: uncomment the .....```

upt_macports/templates/base.Portfile
```diff
@@ -4,16 +4,20 @@ PortSystem          1.0
...
 # supported_archs     noarch
+license             {{ pkg.licenses }}
+maintainers         {example.org:julesverne @GitHub-handle} openmaintainer
```
reneeotten: comment this line out and add TODO at the end:
```# maintainers         {example.org:julesverne @GitHub-handle} openmaintainer # TODO```

**Review by reneeotten — CHANGES_REQUESTED** [review]
a suggestions to follow-up on what @Steap already mentioned, besides that it looks good to go!

**Comment by Korusuke** [comment]
I have done the necessary changes and will update the Portfile over at macports/ports once this PR is merged.

**Comment by reneeotten** [comment]
this looks good to me and it can always be fine-tuned based on input from the MacPorts community later. @Steap, since it's really Portfile/MacPorts related and your suggestions incorporated I'm going to assume that your fine with me merging it ;)

File: upt_macports/templates/base.Portfile [hunk 6]
```diff
@@ -4,16 +4,22 @@ PortSystem          1.0
 {% block portgroup %}
 {% endblock %}
 
+# Portfile generated/maintained using the Universal Packaging Tool (upt)
```
mojca, line 7: I would probably use `maintained` only after we get automatic updates working :)

upt_macports/templates/base.Portfile
```diff
@@ -4,16 +4,22 @@ PortSystem          1.0
...
+# Please adapt the lines with a "TODO" statement and check the guide for more information:
+# https://guide.macports.org/#reference.keywords
 
+# categories-append # TODO
 platforms           darwin
-# uncomment and edit the following line depending on whether the port installs architecture-dependent files
+# TODO: uncomment the following line if no architecture-dependent files are installed, otherwise remove
 # supported_archs     noarch
+license             {{ pkg.licenses }}
+# maintainers         {example.org:julesverne @GitHub-handle} openmaintainer # TODO
```
mojca, line 19: I would also keep a line for `nomaintainer` and let the user decide between the two.
It doesn't serve anyone to force users to assume maintainership when they are not comfortable to do so and not willing to take care of the port later.

upt_macports/templates/base.Portfile
reneeotten: we are optimistic about it ;)

upt_macports/templates/base.Portfile
reneeotten: yes, that's a valid point - how about something like this?
```
# TODO: either add your e-mail address and GitHub handle or set to nomaintainer
# maintainers         {example.org:julesverne @GitHub-handle} openmaintainer
# maintainers         nomaintainer
```

mojca: One further idea. At the end the port would ideally not contain any comments. So maybe the `TODO`s are a bit superfluous. Maybe something along those lines?

```
## If you would like to keep maintaining this port in the future as described in
##     https://guide.macports.org/chunked/project.contributing.html
## please add your contacts (e-mail, github handle), else keep `nomaintainer`.
# maintainers         {example.org:julesverne @github-handle} openmaintainer
# maintainers         nomaintainer
```

And then also ask users in a comment somewhere on top to go through the commented-out sections and check them out (and ideally remove most comments).

PR #18, opened 2019-05-29
Update license conversion

### LINKED ISSUE #14: Update port
The `py-upt-macports` port is currently generating defunct packages, including #11 and printing empty licence name (https://github.com/macports/macports-ports/pull/4489).

Can you please update the `Portfile`?

@rajdeepbharati

Korusuke:
Return unknown if no license found and test for the same

Fixes #14

### DIFF
--- a/upt_macports/tests/test_macports_package.py
+++ b/upt_macports/tests/test_macports_package.py
@@ -10,7 +10,7 @@ class TestMacPortsPackageLicenses(unittest.TestCase):
 
     def test_no_licenses(self):
         self.package.upt_pkg.licenses = []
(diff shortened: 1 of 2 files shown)


reneeotten:
@Korusuke this seems to work as intended, but the return-statement becomes a bit hard to decipher... not sure if that's a real issue but readability does count to a certain point. @Steap what do you think?

Steap:
It is not super elegant, but having an "if/else" may not be that much better. I think it is down to personal preference here.

reneeotten:
okay, let's merge this then

PR #19, opened 2019-05-30
Perl portfile updated

### LINKED ISSUE #17: Sort dependencies in perl (cpan)

Korusuke:
Fixes #17

### DIFF
--- a/upt_macports/templates/perl.Portfile
+++ b/upt_macports/templates/perl.Portfile
@@ -5,16 +5,40 @@ PortGroup           perl5 1.0
 {% endblock %}
 {% block nameversion %}
 perl5.branches      5.26 5.28
-perl5.setup         {{ pkg._pkgname() }} {{ pkg.upt_pkg.version }}
+perl5.setup         {{ pkg._pkgname() }} {{ pkg.upt_pkg.version }}{{ pkg._cpandir() }}
 revision            0
 {% endblock %}
 
 {% block versions %}
+{%- if pkg.upt_pkg.requirements.run or pkg.upt_pkg.requirements.test or pkg.upt_pkg.requirements.build or pkg.upt_pkg.requirements.config %}
 if {${perl5.major} != ""} {
-    depends_lib-append \
-                    {% for i in pkg.upt_pkg.requirements.run %}
-                    port:p${perl5.major}-{{i.name|replace('::','-')|lower}} {% if loop.index != pkg.upt_pkg.requirements.run|length %}\{% endif %}
+    {%- if pkg.upt_pkg.requirements.config or pkg.upt_pkg.requirements.build %}
 
+    depends_build-append \
+                    {% for i in (pkg.upt_pkg.requirements.build + pkg.upt_pkg.requirements.config)|sort(attribute='name')|unique(attribute='name') %}
+                    port:p${perl5.major}-{{i.name|replace('::','-')|lower}}{% if not loop.last %} \
+                    {% endif %}
                     {% endfor %}
+    {% endif %}
+
+    {% if pkg.upt_pkg.requirements.run %}
+
+    depends_lib-append \
+                    {% for i in pkg.upt_pkg.requirements.run|sort(attribute='name') %}
+                    port:p${perl5.major}-{{i.name|replace('::','-')|lower}}{% if not loop.last %} \
+                    {% endif %}
+                    {% endfor %}
+    {% endif %}
+
+    {% if pkg.upt_pkg.requirements.test %}
+
+    depends_test-append \
+                    {% for i in pkg.upt_pkg.requirements.test|sort(attribute='name') %}
+                    port:p${perl5.major}-{{i.name|replace('::','-')|lower}}{% if not loop.last %} \
+                    {% endif %}
+                    {% endfor %}
+    {% endif %}
+
 }
+{% endif %}
 {% endblock %}
--- a/upt_macports/tests/test_perl_package.py
+++ b/upt_macports/tests/test_perl_package.py
@@ -1,0 +1,45 @@
+import unittest
+
+import requests_mock
+
+import upt
+
+from upt_macports.upt_macports import MacPortsPerlPackage
+
+
+class TestMacPortsPerlPackage(unittest.TestCase):
+    def setUp(self):
+        self.package = MacPortsPerlPackage()
+        self.package.upt_pkg = upt.Package('Foo-Bar', '13.37')
+        self.package.upt_pkg.archives = [
+            upt.Archive("https://democpan.org/authors/id/F/FO/FOOBAR/Foo-Bar-13.37.tar.gz")] # noqa
+        self.check_url = "https://cpan.metacpan.org/modules/by-module/Foo/Foo-Bar-13.37.tar.gz" # noqa
+
+    def test_pkgname(self):
+        expected = ['Foo-bar', 'foo-bar', 'Foo-bar', 'foo-bar']
+        names = ['Foo::bar', 'foo::bar', 'Foo-bar', 'foo-bar']
+        for (name, expected_name) in zip(names, expected):
+            self.package.upt_pkg = upt.Package(name, '13.37')
+            self.assertEqual(self.package._pkgname(), expected_name)
+
+    @requests_mock.mock()
+    def test_missing_dist_pos1(self, requests):
+        expected = ' ../../authors/id/F/FO/FOOBAR/'
+        requests.head(self.check_url, status_code=400)
+        self.assertEqual(self.package._cpandir(), expected)
+
+    @requests_mock.mock()
+    def test_distfile_found(self, requests):
+        expected = ''
+        requests.head(self.check_url, status_code=200)
+        self.assertEqual(self.package._cpandir(), expected)
+
+    @requests_mock.mock()
+    def test_missing_distfile(self, requests):
+        expected = ' # could not locate dist file'
+        self.package.upt_pkg.archives = []
+        self.assertEqual(self.package._cpandir(), expected)
+
+
+if __name__ == '__main__':
+    unittest.main()
--- a/upt_macports/upt_macports.py
+++ b/upt_macports/upt_macports.py
@@ -3,6 +3,7 @@ import logging
 import jinja2
 import pkg_resources
 import json
+import requests
 
 
 class MacPortsPackage(object):
@@ -105,7 +106,31 @@ class MacPortsPerlPackage(MacPortsPackage):
 
     @staticmethod
     def _normalized_macports_name(name):
-        return name
+        return name.replace('::', '-')
+
+    def _cpandir(self):
+        pkg = self.upt_pkg
+        # If no archives detected then we cannot locate dist file
+        if not pkg.archives:
+            self.logger.warning('No dist file was found')
+            return ' # could not locate dist file'
+
+        # We start by checking at usual location
+        archive_name = pkg.archives[0].url.split('/')[-1]
+        part_name = pkg.name.replace('::', '-').split('-')[0]
+        check_url = f'https://cpan.metacpan.org/modules/by-module/{part_name}/{archive_name}' # noqa
+        r = requests.head(check_url)
+        if r.status_code == 200:
+            self.logger.info('Dist file found at usual location')
+            return ''
+        else:
+            # Sometimes if it is not available,
+            # then we fallback to alternate location
+            # to be verified by the maintainer
+            fallback_dist = '/'.join(pkg.archives[0].url.split('id/')[1].split('/')[:-1]) # noqa
+            self.logger.info('Dist file was not found at usual location')
+            self.logger.info('Using fallback location for dist file')
+            return f' ../../authors/id/{fallback_dist}/'
 
 
 class MacPortsRubyPackage(MacPortsPackage):
--- a/setup.cfg
+++ b/setup.cfg
@@ -17,6 +17,7 @@ classifiers =
 [options]
 packages = find:
 install_requires =
+    requests
     upt
     jinja2
 include_package_data = true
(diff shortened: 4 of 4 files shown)


upt_macports/templates/perl.Portfile
```diff
@@ -10,11 +10,24 @@ revision            0
 {% endblock %}
 
 {% block versions %}
+{%- if (pkg.upt_pkg.requirements.run|length > 0) or (pkg.upt_pkg.requirements.test|length > 0) %}
```
Steap: Can't we write:

> {% if pkg.upt_pkg.requirements.run or pkg.upt_pkg.requirements.test %}

here?

upt_macports/templates/perl.Portfile
```diff
@@ -10,11 +10,33 @@ revision            0
 {% endblock %}
 
 {% block versions %}
+{%- if pkg.upt_pkg.requirements.run or pkg.upt_pkg.requirements.test or pkg.upt_pkg.requirements.build %}
 if {${perl5.major} != ""} {
+    {%- if pkg.upt_pkg.requirements.build|length > 0 %}
+
+    depends_build-append \
+                    {% for i in pkg.upt_pkg.requirements.build|sort(attribute='name') %}
+                    port:p${perl5.major}-{{i.name|replace('::','-')|lower}} {% if loop.index != pkg.upt_pkg.requirements.build|length %}\{% endif %}
```
reneeotten: can ```{% if loop.index != pkg.upt_pkg.requirements.build|length %}``` not just be ```{% if not loop.last %}```? If so, you use this a few more times below as well.

Steap: You're right. It is also what we do in Python.Portfile iirc.

reneeotten:
Follow-up on the e-mail from this morning with respect to the ```perl5.setup``` line; in the ```cpan2port``` code it says:

"Note that two versions of perl5.setup are generated. The second is for use when the distfile isn't in the customary location (unfortunately happening more and more often on CPAN). In the future, it would be nice to have a command line option to choose between the two, or better yet have it chosen automatically with a couple of quick Web checks."

That gives a clear pointer of why the additional ```../by-authors``` comes from. So I would compare a few ports that have this additional element with ports that don't and see if you from the (full) json metadata can figure out how to automatically determine when the distfile isn't in the "customary location". After that and when Issue #22 is resolved the perl Porftiles should be pretty much "production-ready" or at least ready for testing, correct?

Korusuke:
> Follow-up on the e-mail from this morning with respect to the `perl5.setup` line; in the `cpan2port` code it says:
> 
> "Note that two versions of perl5.setup are generated. The second is for use when the distfile isn't in the customary location (unfortunately happening more and more often on CPAN). In the future, it would be nice to have a command line option to choose between the two, or better yet have it chosen automatically with a couple of quick Web checks."

Following this, I have automated the check and the extra part (/authors) will only be added when needed.

> That gives a clear pointer of why the additional `../by-authors` comes from. So I would compare a few ports that have this additional element with ports that don't and see if you from the (full) json metadata can figure out how to automatically determine when the distfile isn't in the "customary location". After that and when Issue #22 is resolved the perl Porftiles should be pretty much "production-ready" or at least ready for testing, correct?

Still, there's a bit of confusion in the authors argument, following are the variations:

- `../by-authors/id/M/MN/MNOONING/PlRPC/` PlRPC 0.2020
- `../by-authors/id/N/NE/NEILB` Smart-Comments 1.06
- `../../authors/id/A/AD/ADOPTME` PodToHTML 0.09
- `../../authors/id/N/NE/NEILB/` Net-Dict 2.21

here there are two variables :
1. `by-authors` / `authors`
2. the trailing `/`

As per what I tried out on all 50 ports(random chosen), the last format always works.
So I guess, we should go ahead with it.

To get this working, I had to change `upt` as it was not passing the `download_urls`.
From what I see in the commit log of `upt`, @Steap has recently removed `download_urls`.
I guess that would be because of Archives and I am realizing that just now while writing this, so if this is true then I will make the changes accordingly with any other changes that are suggested.

Also, I know that tests for the same will be required and I will try and push that today.

Korusuke:
Had to do a few more changes like only taking unique dependencies from the combined list of build and configure.
Another thing was a trailing space if no author is there, it's kinda fixed now.
Kinda because the template file looks a bit weird without space and the return statement have a starting space.

The link for new diff is: [perldiff3](https://perldiff3.kiteretsu.now.sh)

Korusuke:
Test added although the names of those tests are pretty bad ;)

upt_macports/templates/perl.Portfile
```diff
@@ -5,16 +5,40 @@ PortGroup           perl5 1.0
...
 perl5.branches      5.26 5.28
-perl5.setup         {{ pkg._pkgname() }} {{ pkg.upt_pkg.version }}
+perl5.setup         {{ pkg._pkgname() }} {{ pkg.upt_pkg.version }}{{ pkg._cpandir() }}
 revision            0
 {% endblock %}
 
 {% block versions %}
+{%- if pkg.upt_pkg.requirements.run or pkg.upt_pkg.requirements.test or pkg.upt_pkg.requirements.build %}
 if {${perl5.major} != ""} {
+    {%- if pkg.upt_pkg.requirements.build|length > 0 %}
```
Steap: Remove all the "| length > 0".

upt_macports/templates/perl.Portfile
```diff
@@ -5,16 +5,40 @@ PortGroup           perl5 1.0
...
+
+    depends_build-append \
+                    {% for i in pkg.upt_pkg.requirements.build|sort(attribute='name')|unique(attribute='name') %}
+                    port:p${perl5.major}-{{i.name|replace('::','-')|lower}}{% if not loop.last %} \
```
Steap, line 19: Why not use "_normalized_macports_name" here, since it is a static method? Should "lower" be part of this method as well?

upt_macports/upt_macports.py
```diff
@@ -105,7 +106,23 @@ def _pkgname(self):
...
+            check_url = f'https://cpan.metacpan.org/modules/by-module/{part_name}/{archive_name}' # noqa
+            r = requests.head(check_url)
+            if r.status_code == 200:
+                print(r)
+                return ''
+            else:
+                fallback_dist = '/'.join(pkg.archives[0].url.split('id/')[1].split('/')[:-1]) # noqa
+                if fallback_dist:
+                    return f' ../../authors/id/{fallback_dist}/'
+        return ' # could not locate dist file'
```
Steap: So, this function deserves a big wall of text (or multiple ones) to explain what is happening. Also, I would rather eliminate the weirdest case first, like people do in C. In C, you often see:

```c
void foo()
{
    if (something_weird())
        return;

    if (something_weird_too())
        return;

    /* And now, for the real stuff */
    ...
}
```

You get into the "main" part of the function with less context to remember, and you gain a level of indentation. So, here, we could write:

```python
    if not pkg.archives:
        return ' # No dist file'

    # We start by checking the usual URL
    blabla
    if r:
        return ''
    else:
        # Sometimes, it is not available, in which case we fall back to the following:
        return f'...'
```

What do you think?

Korusuke: Yeah agree this makes more sense, will push the changes in some time.

upt_macports/templates/perl.Portfile
Korusuke: I don't think we can keep lower as part of "_normalized_macports_name" as for "perl5.setup" we need the normal version.

Also writing  `{{i.name|replace('::','-')|lower}}` looks a bit cleaner then
`{{pkg._normalized_macports_name(i.name)|lower}}` I guess.

what do you think?

upt_macports/tests/test_perl_package.py
```diff
@@ -0,0 +1,45 @@
...
+
+from upt_macports.upt_macports import MacPortsPerlPackage
+
+
+class TestMacPortsPerlPackage(unittest.TestCase):
+    def setUp(self):
+        self.package = MacPortsPerlPackage()
+        self.package.upt_pkg = upt.Package('Foo-Bar', '13.37')
+        self.package.upt_pkg.archives = [upt.Archive(
+          "https://democpan.org/authors/id/F/FO/FOOBAR/Foo-Bar-13.37.tar.gz")]
```
Steap: Usually, one would write:

```python
        self.package.upt_pkg.archives = [
            upt.Archive(...)
        ]
```

Steap:
upt-cpan 0.5 adds support for "config" and "build". I know I said I'd merge these two into "build", but I ended up going for two separate fields. Sorry!

Korusuke:
> upt-cpan 0.5 adds support for "config" and "build". I know I said I'd merge these two into "build", but I ended up going for two separate fields. Sorry!

No issues at all, I had tried that way too.
Updated the PR :)

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
reneeotten: these type of changes shouldn't be here... please add ```.vscode``` to ```.gitignore```

upt_macports/templates/perl.Portfile
```diff
@@ -5,16 +5,40 @@ PortGroup           perl5 1.0
...
 revision            0
 {% endblock %}
 
 {% block versions %}
+{%- if pkg.upt_pkg.requirements.run or pkg.upt_pkg.requirements.test or pkg.upt_pkg.requirements.build or pkg.upt_pkg.requirements.config %}
 if {${perl5.major} != ""} {
+    {%- if pkg.upt_pkg.requirements.config or pkg.upt_pkg.requirements.build %}
+
+    depends_build-append \
+                    {% for i in (pkg.upt_pkg.requirements.build + pkg.upt_pkg.requirements.config) |sort(attribute='name')|unique(attribute='name') %}
```
reneeotten: should be a space between the ```|``` both for the "sort" and "unique" (or at least be consistent on both sides)

upt_macports/templates/perl.Portfile
reneeotten: this changed was made, anything else here @Steap - otherwise just press "Resolve"

upt_macports/templates/perl.Portfile
reneeotten: Yes, there is indeed the need for a different name for the ```perl5.setup``` line and the dependencies list... I am sure there are different options, but what it's right now seems fine to me. It doesn't really hurt the readability of the template, so I would just leave it for now in the interest of merging this soon...

reneeotten:
@Steap is there anything else left to do from you earlier review comments. I looked at the updated comparison of the Perl ports from @Korusuke and it seems to me that with the current changes ```upt``` does a very good job.

I would be in favor of merging this soon and possibly make small changes to the template or code later (we already established earlier that the white-space control is a bit mysterious, so if the template itself looks a bit strange, but the final output is correct - that's fine with me)!

upt_macports/templates/perl.Portfile
mojca: My personal preference would be to call a function rather than regexping on spot, in particular for cases when suddenly one decides to change the convention (like change capitalisation), but I leave it to the mentors.

I would in fact use `pkg.whatever_the_function_name_is(i.name)` to replace the full `port:p${perl5.major}-{{i.name|replace('::','-')|lower}}`.

reneeotten: as I said that is a possibility, it would need another function like ```pkg._macports_dependency_name``` that does an additional ```.lower()``` on ```pkg._normalized_macports_name```

Steap: @mojca: for now, we can probably keep it like that. But in the future, we could move this whole section back to the base template, and have a custom filter do the "formatting" of the package name.

.vscode/settings.json
Steap: In a separate commit :)

Steap:
Well, there is still that vscode file, and my comment on  upt_macports/tests/test_perl_package.py .

reneeotten:
> Well, there is still that vscode file, and my comment on upt_macports/tests/test_perl_package.py .

Right. Your request with respect to the tests is only to change the formatting from:
```
        self.package.upt_pkg.archives = [upt.Archive(
            "https://democpan.org/authors/id/F/FO/FOOBAR/Foo-Bar-13.37.tar.gz")]
```

to 
```
        self.package.upt_pkg.archives = [
             upt.Archive("https://democpan.org/authors/id/F/FO/FOOBAR/Foo-Bar-13.37.tar.gz")]
```

correct?

upt_macports/templates/perl.Portfile
reneeotten: I *think* I see now what @mojca means: we should add a function in the backend(s) that has ```name``` as argument and returns the complete dependency line. Something like (as pseudo-code):

```
@staticmethod
def _macports_dependency(name):
    dep_type = 'p${perl5.major}'
    dep_name = pkg._normalized_macports_name(name).lower()
    return f'port:{dep_type}-{dep_name}'
```

I do like that suggestion and we should probably do that for every "frontend" that is available. @Korusuke can you please add that to your to-do list (it can probably go into one PR with separate commits for the different Python/Perl/Ruby frontends)?

```

Korusuke: Yeah sure, It can also check for if ports are already present or not on MacPorts tree and decide the name accordingly in case of python ports.
We can implement this during recursive feature addition.

Steap: Let's do that later :)

Steap:
So, @reneeotten , do you feel good about this one? To me, it seems ready to be merged.

reneeotten:
> So, @reneeotten , do you feel good about this one? To me, it seems ready to be merged.

Yes I do - in it goes!

PR #20, opened 2019-05-30
Add Ruby portfile

### LINKED ISSUE #16: RubyGems template
Template for RubyGems frontend is missing and needs to be added.

Korusuke:
Closes #16

### DIFF
--- a/upt_macports/templates/ruby.Portfile
+++ b/upt_macports/templates/ruby.Portfile
@@ -1,0 +1,34 @@
+{% extends 'base.Portfile' %}
+
+{% block portgroup %}
+PortGroup           ruby 1.0
+{% endblock %}
+{% block nameversion %}
+ruby.setup          {{ pkg._pkgname() }} {{ pkg.upt_pkg.version }} gem {} rubygems
+ruby.branches       2.5 2.6
+revision            0
+{% endblock %}
+
+{% block versions %}
+{% if pkg.upt_pkg.requirements.run or pkg.upt_pkg.requirements.test %}
+if {${subport} ne ${name}} {
+    {% if pkg.upt_pkg.requirements.run %}
+    depends_lib-append \
+                    {% for i in pkg.upt_pkg.requirements.run|sort(attribute='name') %}
+                    port:rb${ruby.version}-{{i.name}}{% if not loop.last %} \
+                    {% endif %}
+                    {% endfor %}
+    {% endif %}
+
+
+    {% if pkg.upt_pkg.requirements.test %}
+    depends_test-append \
+                    {% for i in pkg.upt_pkg.requirements.test|sort(attribute='name') %}
+                    port:rb${ruby.version}-{{i.name}}{% if not loop.last %} \
+                    {% endif %}
+                    {% endfor %}
+    {% endif %}
+
+}
+{% endif %}
+{% endblock %}
--- a/upt_macports/tests/test_ruby_package.py
+++ b/upt_macports/tests/test_ruby_package.py
@@ -1,0 +1,22 @@
+import unittest
+
+import upt
+
+from upt_macports.upt_macports import MacPortsRubyPackage
+
+
+class TestMacPortsRubyPackage(unittest.TestCase):
+    def setUp(self):
+        self.package = MacPortsRubyPackage()
+        self.package.upt_pkg = upt.Package('test-pkg', '13.37')
+
+    def test_pkgname(self):
+        expected = ['Foo', 'foo', 'Foo-bar', 'foo-bar']
+        names = ['Foo', 'foo', 'Foo-bar', 'foo-bar']
+        for (name, expected_name) in zip(names, expected):
+            self.package.upt_pkg = upt.Package(name, '13.37')
+            self.assertEqual(self.package._pkgname(), expected_name)
+
+
+if __name__ == '__main__':
+    unittest.main()
--- a/upt_macports/upt_macports.py
+++ b/upt_macports/upt_macports.py
@@ -111,6 +111,14 @@ class MacPortsPerlPackage(MacPortsPackage):
 class MacPortsRubyPackage(MacPortsPackage):
     template = 'ruby.Portfile'
 
+    def _pkgname(self):
+        macports_name = self._normalized_macports_name(self.upt_pkg.name)
+        return macports_name
+
+    @staticmethod
+    def _normalized_macports_name(name):
+        return name
+
 
 class MacPortsBackend(upt.Backend):
     name = 'macports'
--- a/upt_macports/templates/base.Portfile
+++ b/upt_macports/templates/base.Portfile
@@ -12,8 +12,10 @@ PortSystem          1.0
 
 # categories-append
 platforms           darwin
+{% if pkg.upt_pkg.frontend != 'rubygems' -%}
 ## uncomment the following line if no architecture-dependent files are installed, otherwise remove
 # supported_archs     noarch
+{% endif %}
 license             {{ pkg.licenses }}
 ## If you would like to keep maintaining this port in the future as described in
 ## https://guide.macports.org/chunked/project.contributing.html then
@@ -28,9 +30,9 @@ long_description    ${description}
 {% block dist_info %}
 {% endblock -%}
 
-{% if pkg.archive_type == 'unknown' -%}
+{% if pkg.archive_type == 'unknown' and pkg.upt_pkg.frontend != 'rubygems' -%}
 # Detection of archive type failed
-{% elif pkg.archive_type != 'gz' %}
+{% elif pkg.archive_type != 'gz' and pkg.upt_pkg.frontend != 'rubygems' %}
 {{ "use_%-15s yes"|format(pkg.archive_type) }}
 {% endif %}
 


upt_macports/upt_macports.py
```diff
@@ -111,6 +111,14 @@ def _normalized_macports_name(name):
...
     template = 'ruby.Portfile'
 
+    def _pkgname(self):
+        macports_name = self._normalized_macports_name(self.upt_pkg.name)
+        return macports_name
+
+    @staticmethod
+    def _normalized_macports_name(name):
+        return name
+
```
Steap, line 121: Isn't this a bit over-engineered for now? This could be:

```python
def _pkgname(self):
     return self.upt_pkg.name
```

Also, even if this is simple, it needs unit tests.

Steap:
Overall, I think @reneeotten is more qualified than me to comment on this PR. I see you've also talked to @mojca about this.

upt_macports/tests/test_ruby_package.py
```diff
@@ -0,0 +1,22 @@
...
+        self.package = MacPortsRubyPackage()
+        self.package.upt_pkg = upt.Package('test-pkg', '13.37')
+
+    def test_pkgname(self):
+        expected = ['Foo', 'foo', 'Foo-bar', 'foo-bar']
+        names = ['Foo', 'foo', 'Foo-bar', 'foo-bar']
+        for i in range(len(names)):
+            self.package.upt_pkg = upt.Package(names[i], '13.37')
+            self.assertEqual(self.package._pkgname(), expected[i])
+
```
Steap, line 19: "for i in range(len(...))" is not really Pythonic.

Why not something like:

```python
for (name, expected_name) in zip(names, expected):
    ...
```

Steap:
Please squash the commits: code should come with tests.

upt_macports/templates/ruby.Portfile
```diff
@@ -0,0 +1,27 @@
+{% extends 'base.Portfile' %}
+
+{% block portgroup %}
+PortGroup           ruby 1.0
+{% endblock %}
+{% block nameversion %}
```
reneeotten, line 6: similarly to the Perl template, here you should add: ```ruby.branches       ???```, where ??? should of course be the version or versions. At least for now, in the testing phase, I would only add the latest two version so ```2.5 2.6```. 

You should probably take a look at [the Ruby PortGroup](https://github.com/macports/macports-ports/blob/master/_resources/port1.0/group/ruby-1.0.tcl) to see if there is anything else that should be defined in the template. It supports two use-cases, I would recommend to go with option 1 so that it's similar to the usage for Perl ports. Likely, the PortGroup will need some revamping to support these newer versions but that's outside of the scope of the ```upt``` project. Let's see if we can port some Ruby packages and then adapt the PortGroup where necessary .

upt_macports/templates/ruby.Portfile
```diff
@@ -0,0 +1,27 @@
...
+ruby.setup          {{ pkg._pkgname() }} {{ pkg.upt_pkg.version }} gem {} rubygems
+revision            0
+{% endblock %}
+
+{% block versions %}
+{%- if pkg.upt_pkg.requirements.run|length > 0 %}
```
reneeotten: using the ```ruby.branches``` approach it will generate subports similarly to the situation for Perl/Python, so for example ```rb25-<name>``` and ```rb26-<name>```. I'd imagine that we use the same ```if {${subport} ne ${name}} {``` block as in the Python ports here as well.

reneeotten:
A few other things I noticed when looking at your [comparison](https://rubydiff.kiteretsu.now.sh/rb-actionpack.html):

* there is never a ```homepage``` field in the Portfile, that should be corrected
* if the ```type``` is ```gem``` we will always get a ```# Detection of archive type failed``` in the template since the extension is ```.gem.```. That's expected and should not result in a ```# Detection of archive type failed``` in the Portfile.
* aren't Ruby gems always "noarch"? This is related to your statement/question in an earlier e-mail "upt-rubygems doesn't have the archive feature as of now, as I had to add that for this to work." It shouldn't need this and the MacPorts backend should handle this gracefully.
* should we be adding "development dependencies" (see [here](https://guides.rubygems.org/specification-reference/#add_development_dependency)); the way I read it these are actually not needed for running but only if you want to develop/make changes to the gem.
* is ```upt-rubygems``` also parsing the ```requirements``` [field](https://guides.rubygems.org/specification-reference) that can be used to declare dependencies external to RubyGems?/#requirements? If not, it would be nice to add this and list the external dependencies in the Portfile for the maintainer to verify/add correctly.
* the ```rb19-<name>``` ports should probably be renamed
* drop all EOL versions of Ruby (see [here](https://www.ruby-lang.org/en/downloads/branches/)). This is something we should probably bring up on the mailinglist, but since we pretty much start from scratch as current Ruby support is mostly lacking, I think it would make sense to rework the PortGroup and Portfiles such that we support only *current* Ruby versions and don't make all kind of exceptions for legacy versions.

Again, I have never used Ruby so I am not 100% sure about this... but this is my current interpretation. @Korusuke I suppose that you read through the [RubyGems Guides](https://guides.rubygems.org/), if not please do so ;)

upt_macports/templates/ruby.Portfile
```diff
@@ -0,0 +1,27 @@
+{% extends 'base.Portfile' %}
+
+{% block portgroup %}
+PortGroup           ruby 1.0
+{% endblock %}
+{% block nameversion %}
+ruby.setup          {{ pkg._pkgname() }} {{ pkg.upt_pkg.version }} gem {} rubygems
```
reneeotten, line 7: based on a quick glance through the PortGroup I don't really understand the need for the ```{}```.... but since it's present in all current Portfiles there must be a reason, can you please check?

upt_macports/templates/ruby.Portfile
Korusuke: Yeah I saw the Ruby port group, but seeing the existing Ruby Ports I was not sure about `ruby.branches` so did not add them as only one port is currently using that.
Also as far as I can see we cannot automate the detection of supported version as that information is not available in the json unlike python(even if this is unreliable), but we can sure hardcode it for now.

But as almost none of the ports use it, the question is, is it really needed?

Korusuke:
> * if the `type` is `gem` we will always get a `# Detection of archive type failed` in the template since the extension is `.gem.`. That's expected and should not result in a `# Detection of archive type failed` in the Portfile.

Ok will add a condition for this.

> * aren't Ruby gems always "noarch"? This is related to your statement/question in an earlier e-mail "upt-rubygems doesn't have the archive feature as of now, as I had to add that for this to work." It shouldn't need this and the MacPorts backend should handle this gracefully.

Not sure but I guess we will need the `archive` feature in `upt-rubygems` so as to calculate the checksums, I have done the changes in my local copy.
Once you confirm that it's needed I will raise an issue on upt-rubygems.

> * should we be adding "development dependencies" (see [here](https://guides.rubygems.org/specification-reference/#add_development_dependency)); the way I read it these are actually not needed for running but only if you want to develop/make changes to the gem.

Yeah as the frontend is already providing with the dev. dependencies, we can add those.

> * is `upt-rubygems` also parsing the `requirements` [field](https://guides.rubygems.org/specification-reference) that can be used to declare dependencies external to RubyGems?/#requirements? If not, it would be nice to add this and list the external dependencies in the Portfile for the maintainer to verify/add correctly.

This is something that not many package lists(I am still looking for one that has this) and also secondly how would we be writing this cause as per the example it's things like "good graphics card".

> * the `rb19-<name>` ports should probably be renamed

Yes we can do this once the update feature is working.

upt_macports/templates/ruby.Portfile
reneeotten: Yes, I think so. Just as for Python and Perl we want to be able to support different versions, right now it likely defaults to ```ruby.default_branch```. Even if we only add ```2.6``` for new ports (similarly as we only add ```37``` for ```python.versions``` in the Python template), it's still useful to add in the Portfile so that if a new version comes out we can easily add it.

reneeotten:
> > * aren't Ruby gems always "noarch"? This is related to your statement/question in an earlier e-mail "upt-rubygems doesn't have the archive feature as of now, as I had to add that for this to work." It shouldn't need this and the MacPorts backend should handle this gracefully.
> 
> Not sure but I guess we will need the `archive` feature in `upt-rubygems` so as to calculate the checksums, I have done the changes in my local copy. Once you confirm that it's needed I will raise an issue on upt-rubygems.

sorry, this are probably indeed two different things:
1. please check whether RubyGems are always add ```noarch```, in that case you could add the ```supported_archs noarch``` always to the Portfile, or we can set it in the PortGroup.
2. we certainly want to add ```checksums```, so if that requires an "archive feature" in the frontend please do add that!

> > * should we be adding "development dependencies" (see [here](https://guides.rubygems.org/specification-reference/#add_development_dependency)); the way I read it these are actually not needed for running but only if you want to develop/make changes to the gem.
> 
> Yeah as the frontend is already providing with the dev. dependencies, we can add those.

Right, but as I said you should investigate if they are actually needed for running/installing the gem...

> > * is `upt-rubygems` also parsing the `requirements` [field](https://guides.rubygems.org/specification-reference) that can be used to declare dependencies external to RubyGems?/#requirements? If not, it would be nice to add this and list the external dependencies in the Portfile for the maintainer to verify/add correctly.
> 
> This is something that not many package lists(I am still looking for one that has this) and also secondly how would we be writing this cause as per the example it's things like "good graphics card".

Again, I have no experience with Ruby but from skimming the documentation I thought this might be useful (not the "good graphics card" of course...). I haven't looked if there are actually packages around that use this, but you could take a closer look at that. But I agree it shouldn't be high priority.

Korusuke:
> sorry, this are probably indeed two different things:
> 
> 1. please check whether RubyGems are always add `noarch`, in that case you could add the `supported_archs noarch` always to the Portfile, or we can set it in the PortGroup.

I checked the Portfile and as the portfile sets noarch, I think we don't need that field.

> 2. we certainly want to add `checksums`, so if that requires an "archive feature" in the frontend please do add that!

I have created an [issue](https://framagit.org/upt/upt-rubygems/issues/1) at upt-rubygems for the same

upt_macports/upt_macports.py
reneeotten: I think the tests are present now. I agree that this function is not really needed right now, but for the sake of uniformity between code for different frontends, I would be fine with leaving it. If in the end it turns out that we really don't need any conversion of the name we can completely remove it.

upt_macports/templates/ruby.Portfile
```diff
@@ -0,0 +1,34 @@
...
+ruby.branches       2.5 2.6
+revision            0
+{% endblock %}
+
+{% block versions %}
+{% if pkg.upt_pkg.requirements.run or pkg.upt_pkg.requirements.test %}
+if {${subport} ne ${name}} {
+    {% if pkg.upt_pkg.requirements.run|length > 0 %}
```
reneeotten: the "length" part should be removed everywhere in the ruby.Portfile template as well (just as you did for the Perl template)

reneeotten:
okay, a few more tiny details. @Steap can you check whether there is anything left to do - otherwise let's merge this soon!

upt_macports/templates/ruby.Portfile
```diff
@@ -0,0 +1,34 @@
...
+{% endblock %}
+
+{% block versions %}
+{% if pkg.upt_pkg.requirements.run or pkg.upt_pkg.requirements.test %}
+if {${subport} ne ${name}} {
+    {% if pkg.upt_pkg.requirements.run %}
+    depends_lib-append \
+                    {% for i in pkg.upt_pkg.requirements.run|sort(attribute='name') %}
+                    port:rb${ruby.version}-{{i.name}}{% if not loop.last %} \
+                    {% endif %}
```
reneeotten, line 19: please move this ``` {% endif %}``` to the previous line to close the ```if``` opening statement

upt_macports/templates/ruby.Portfile
```diff
@@ -0,0 +1,34 @@
...
+                    {% endfor %}
+    {% endif %}
+
+
+    {% if pkg.upt_pkg.requirements.test %}
+    depends_test-append \
+                    {% for i in pkg.upt_pkg.requirements.test|sort(attribute='name') %}
+                    port:rb${ruby.version}-{{i.name}}{% if not loop.last %} \
+                    {% endif %}
```
reneeotten, line 28: please move this ``` {% endif %}``` to the previous line to close the ```if``` opening statement

reneeotten (commented):
sorry @Korusuke can you please take care of the two formatting suggestions in the ruby.Portfile template?

reneeotten:
> I have created an [issue](https://framagit.org/upt/upt-rubygems/issues/1) at upt-rubygems for the same

Quick other question: what happens right now, do we not get a checksum at all? It seems from your comparison of upt-generated and original Portfiles for ruby that the checksums are already present; it looks like it's calculated from the file if it cannot be obtained from the "archive". So we still get the checksums field correctly if we merge this now, irrespective of the possible upcoming changes in ```upt-rubygems```, correct?

Korusuke:
> > I have created an [issue](https://framagit.org/upt/upt-rubygems/issues/1) at upt-rubygems for the same
> 
> Quick other question: what happens right now, do we not get a checksum at all? It seems from your comparison of upt-generated and original Portfiles for ruby that the checksums are already present; it looks like it's calculated from the file if it cannot be obtained from the "archive". So we still get the checksums field correctly if we merge this now, irrespective of the possible upcoming changes in `upt-rubygems`, correct?

No, we do not get any checksums from `upt-rubygems` currently.
The checksums are present in the diff files because I used the local version of `upt-rubygems` in which I added the checksum feature.

reneeotten:
thanks for the changes of the ```{% endif %}```, but now you left an empty line before the ```{% endfor %}```, that's not needed.

I know that you opened an Issue already on ```upt-rubygems```, but what about just submitting a PR if you have the required changes already? @Steap this seems quite straightfoward to add, can we do that quickly so that if we merge this PR the checksums will be available?

Korusuke:
> thanks for the changes of the `{% endif %}`, but now you left an empty line before the `{% endfor %}`, that's not needed.

It is needed, this is another example of weird jinja styling :) 
This was the initial reason of keeping "endif" on next line.

> I know that you opened an Issue already on `upt-rubygems`, but what about just submitting a PR if you have the required changes already? @Steap this seems quite straightfoward to add, can we do that quickly so that if we merge this PR the checksums will be available?

Yeah sure.

reneeotten:
> > thanks for the changes of the `{% endif %}`, but now you left an empty line before the `{% endfor %}`, that's not needed.
> 
> It is needed, this is another example of weird jinja styling :)
> This was the initial reason of keeping "endif" on next line.

well... in that case please change it back - sorry... I don't like this empty-line even less ;) I guess this means that at some point we will need to study the template stuff a bit more and ideally get this a bit more WYSIWYG. I *think* we might need to tweak some environment settings in the ```_render_makefile_template``` function in ```upt_macports.py```; but let's not worry about that too much now...

Korusuke:
Sorry till now I haven't been able to create a PR for upt-rubygems due to crappy internet cause of rains, will do it soon

Steap:
@Korusuke  Don't worry,  I'll take care of it :)

reneeotten:
> @Korusuke Don't worry, I'll take care of it :)

Sure, should we wait with merging this PR until you have released an updated ```upt-rubygems```?

reneeotten:
ping @Steap - to move this forward I'd be in favor of merging this now. Once you have updated ```upt-rubygems``` it will give the checksums as well. If you agree please merge or let us know otherwise!

Steap:
I have just released upt 0.8 and upt-rubygems 0.2!

reneeotten:
> I have just released upt 0.8 and upt-rubygems 0.2!

Cool! I am going to merge this now, then it's ready for some testing and we can fine-tune from there.

Issue #15, opened 2019-05-28
setuptools


Korusuke:
In some cases `setuptools` is returned both in build and lib which should not be there.

Test ports:
py-zope-deprecation (pypi_name : zope-deprecation)
py-buildbot-pkg (pypi_name : buildbot-pkg)
py-fs (pypi_name : fs)

mojca:
Note that some ports might in fact need this at runtime, and it's valid to have it both for runtime and build, it's just strange to have it for build + lib as that's overlapping.

But I assume that most of the python dependencies are in fact runtime dependencies rather than libraries being linked against, so it might as well be a nomenclature issue.

reneeotten:
I think we could/should do a few things to alert the maintainer to double check:
1) add a note stating that (a) we add ```setuptools``` by default to ```depends_build``` since PyPI doesn't provide the information and it's actually most often -but not always- true, (b) that if the package installs a ```console_script``` it should instead be added to ```depends_lib```. Right now we cannot do much more I think and it's up to the maintainer to make sure.
2) if upt tells us that it's a ```depends_lib``` dependency we should not add it as build-dependency (but again the maintainer should check if it's actually true).

jmroot:
> (b) that if the package installs a console_script it should instead be added to depends_lib

Using entry_points will also create a runtime dependency on pkg_resources.

reneeotten:
thanks Josh, that's indeed true and should be included in the comment. @Steap do you think my suggestions above make sense or is there anything else to consider? @Korusuke you could probably draft a PR to resolve this?

For example, if the ```json``` data from PyPI doesn't provide this, could we just parse the setup.py/setup.cfg file(s) and check for the occurrence of ```setuptools``` being imported and the existence of ```console_script``` or ```entry_points``` to get the information with respect to build/lib-dependencies correct?

Steap:
I'm not sure whether 1)b) can easily be detected.

reneeotten:
@Korusuke okay, so short-term we can add a comment for in the python.Portfile regarding ```py-setuptools``` and remove the ```depends_build-append``` in the template when ```upt``` detects it as ```depends_lib```. We can later see if we want to parse the setup.py/setup.cfg (could be done in the MacPorts backend).

PR #21, opened 2019-05-31
Update comments and add setuptools comment

Korusuke:


upt_macports/templates/base.Portfile
```diff
@@ -4,19 +4,21 @@ PortSystem          1.0
 {% block portgroup %}
 {% endblock %}
 
-# Portfile generated/maintained using the Universal Packaging Tool (upt)
+## Portfile generated/maintained using the Universal Packaging Tool (upt)
+## Please check out the commented statements and check the guide for more information:
```
reneeotten: perhaps "...and refer to the guide...."

upt_macports/templates/base.Portfile
```diff
@@ -4,19 +4,21 @@ PortSystem          1.0
...
+##     https://guide.macports.org/chunked/project.contributing.html
```
reneeotten: this link is not correct here, which you could have seen if you opened it. @mojca was referring to this for the "maintainers" field. Here the original link (https://guide.macports.org/#reference.keywords) should stay.

upt_macports/templates/base.Portfile
```diff
@@ -4,19 +4,21 @@ PortSystem          1.0
...
-
-# categories-append # TODO
+# categories-append 
 platforms           darwin
-# TODO: uncomment the following line if no architecture-dependent files are installed, otherwise remove
+## uncomment the following line if no architecture-dependent files are installed, otherwise remove
 # supported_archs     noarch
 license             {{ pkg.licenses }}
-# maintainers         {example.org:julesverne @GitHub-handle} openmaintainer # TODO
+## If you would like to keep maintaining this port in the future as described in
```
reneeotten, line 18: here the link (https://guide.macports.org/chunked/project.contributing.html) should be added

upt_macports/templates/base.Portfile
```diff
@@ -4,19 +4,21 @@ PortSystem          1.0
...
+## please add your contacts (e-mail, github handle), else keep `nomaintainer`.
```
reneeotten: perhaps "## please add your contact information (e-mail and GitHub handle), otherwise keep the `nomaintainer` line."

upt_macports/templates/python.Portfile
```diff
@@ -18,6 +18,7 @@ distname            ${python.rootname}-${version}
 {% block versions %}
 python.versions     37
 
+## setuptools dependency is added by default and should be double checked
```
Steap: Maybe move this comment right before line 24, so that it is close to the piece of code it refers to?

reneeotten: this is not very clear or complete (see the Issue #15)... perhaps something like:
```
## setuptools is added by default as a build-dependency since this is often the case.
## However, if the package installs a "console_script" or uses "entry_points", setuptools
## should be added as a lib-dependency instead. Please verify from the setup.{py/cfg}
## file that setuptools is indeed used and that the dependency-type is correct.
```

and indeed move it to be just above the depedency.

reneeotten (changes requested):
the PR contains all the parts from the e-mail and GitHub conversations...but the details are important. A few suggestions to clarify and/or make it more complete/accurate.

upt_macports/templates/base.Portfile
Korusuke: So if we are keeping both the options, should we do something like keep the `nomaintainer` line uncommented and the other as option?
so it would be like this: 

```
## If you would like to keep maintaining this port in the future as described in
## https://guide.macports.org/chunked/project.contributing.html then
## please add your contacts (e-mail, github handle), otherwise keep the 'nomaintainer' line
# maintainers         {example.org:julesverne @github-handle} openmaintainer
maintainers         nomaintainer
```

upt_macports/templates/python.Portfile
Korusuke: I agree that my initial comment is not complete as I am not so sure about adding so much in comments as this makes it hard to see the actual useful part.
So just an idea, maybe we can add all this in the output before the portfile or just add it in the Readme.
As having it all in the Portfile looks a bit messy and as @mojca said at the end most of the comments should be deleted by the maintainer.
So adding this means more work for the maintainers although I would agree that deleting comments is not time-consuming but I think it could be avoided.

upt_macports/templates/base.Portfile
Korusuke: Sorry, I did open the link but I thought about it in a different way.

upt_macports/templates/base.Portfile
reneeotten: yes, that's indeed a better idea - otherwise it will not generate a "valid" Portfile and ```port lint``` will complain about it. (As a small nitpick, please attempt to keep the length of the comment line as equal as possible, and I still prefer "contact information" over "contacts", but that's personal preference).

upt_macports/templates/python.Portfile
reneeotten: I think it's important to be complete here and in my opinion the whole comment if useful ;) Again, this comments will/should only be added when generating a new Portfile. It's not that hard to read through four lines of comments and know what you're supposed to do as a maintainer (also, removing 4 lines isn't much harder than removing 1).

I am not sure what you mean by "add all this in the output before the portfile"? I am against the idea of adding stuff into a Readme file, despite the name nobody is actually going to read that and comments should be close to where they actually apply.

Korusuke: > I am not sure what you mean by "add all this in the output before the portfile"? I am against the idea of adding stuff into a Readme file, despite the name nobody is actually going to read that and comments should be close to where they actually apply.

Haha that's true no one reads the "Readme".
I meant that we can get rid of the whole comment and not even keep one line as it was in my previous commit and instead when we run upt it will print the "disclaimer/info" comments before the actual upt file(this will ofc be gone once we output to file). But I see that then it would be harder to get that which comment is referring to which line.

upt_macports/templates/base.Portfile
```diff
@@ -4,19 +4,23 @@ PortSystem          1.0
...
+# categories-append
 platforms           darwin
-# TODO: uncomment the following line if no architecture-dependent files are installed, otherwise remove
+## uncomment the following line if no architecture-dependent files are installed, otherwise remove
 # supported_archs     noarch
 license             {{ pkg.licenses }}
-# maintainers         {example.org:julesverne @GitHub-handle} openmaintainer # TODO
+## If you would like to keep maintaining this port in the future as described in
+## https://guide.macports.org/chunked/project.contributing.html then
+## please add your contact infromation (e-mail, github handle)
```
reneeotten: @Korusuke, if you correct the typo (and perhaps change the capitalization and add punctuation) here and change the PR from "Draft" to "Ready" (of whatever it's called), I think we it should be good to go! 

```
## please add your contact information (e-mail and GitHub handle), 
```

PR #24, opened 2019-06-03
Add python name test

Korusuke:
Once this and the other tests of perl and ruby are merged the coverage should reach about 85%.

### DIFF
--- a/upt_macports/tests/test_python_package.py
+++ b/upt_macports/tests/test_python_package.py
@@ -1,0 +1,22 @@
+import unittest
+
+import upt
+
+from upt_macports.upt_macports import MacPortsPythonPackage
+
+
+class TestMacPortsPythonPackage(unittest.TestCase):
+    def setUp(self):
+        self.package = MacPortsPythonPackage()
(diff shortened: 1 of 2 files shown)


reneeotten:
let's first make sure this is indeed the naming convention we decide on before adding tests ;) I have just commented in the Issue. Let's see if Josh or Ryan respond (they will know if there are implicit naming conventions) and take it from there.

reneeotten:
okay, moving forward we'll use the naming ```py-<upstream>``` including capitalization; that certainly makes life easier. We might need to deal with some flexibility for the time being, but let's get to that when working later on checking for//recursive addition of dependencies...

Please rephrase the commit message so that it's clear that we use "py-<upstream>" as MacPorts package name (the tests are implied) and then I will merge.

Korusuke:
Should we just revert this merge and do the changes here?

reneeotten:
> Should we just revert this merge and do the changes here?

If it's up to me: no - I am not in favor of reverting already merged PRs.

PR #25, opened 2019-06-07
Add license logging

Korusuke:
This will only log when upt detected the license but the conversion failed and not when upt itself failed to detect the license.

### DIFF
--- a/upt_macports/upt_macports.py
+++ b/upt_macports/upt_macports.py
@@ -31,8 +31,25 @@ class MacPortsPackage(object):
         with open(filepath) as f:
             spdx2macports = json.loads(f.read())
 
-        return ' '.join([spdx2macports.get(license.spdx_identifier, 'unknown')
-                        for license in self.upt_pkg.licenses]) or 'unknown'
+        if not self.upt_pkg.licenses:
+            self.logger.info('No license found')
+            return 'unknown'
+        licenses = []
+        for license in self.upt_pkg.licenses:
+            try:
+                if license.spdx_identifier == 'unknown':
+                    port_license = 'unknown'
+                    self.logger.warning(f'upt failed to detect license')
+                else:
+                    port_license = spdx2macports[license.spdx_identifier]
+                    self.logger.info(f'Found license {port_license}')
+                licenses.append(port_license)
+            except KeyError:
+                err = f'MacPorts license unknown for {license.spdx_identifier}'
+                licenses.append('unknown # ' + err)
+                self.logger.error(err)
+                self.logger.info('Please report the error at https://github.com/macports/upt-macports') # noqa
+        return ' '.join(licenses)
 
     def _depends(self, phase):
         return self.upt_pkg.requirements.get(phase, [])
--- a/upt_macports/tests/test_macports_package.py
+++ b/upt_macports/tests/test_macports_package.py
@@ -1,6 +1,8 @@
 import unittest
 import upt
-from upt_macports.upt_macports import MacPortsPackage
+from upt_macports.upt_macports import MacPortsPackage, logging
+from unittest import mock
+from io import StringIO
 
 
 class TestMacPortsPackageLicenses(unittest.TestCase):
@@ -18,6 +20,11 @@ class TestMacPortsPackageLicenses(unittest.TestCase):
         expected = 'BSD'
         self.assertEqual(self.package.licenses, expected)
 
+    def test_unknown_license(self):
+        self.package.upt_pkg.licenses = [upt.licenses.ZlibLicense()]
+        expected = 'unknown # MacPorts license unknown for zlib'
+        self.assertEqual(self.package.licenses, expected)
+
     def test_bad_license(self):
         self.package.upt_pkg.licenses = [upt.licenses.UnknownLicense()]
         expected = 'unknown'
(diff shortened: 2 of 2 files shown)


upt_macports/upt_macports.py
```diff
@@ -30,9 +30,15 @@ def licenses(self):
...
         with open(filepath) as f:
             spdx2macports = json.loads(f.read())
-
-        return ' '.join([spdx2macports.get(license.spdx_identifier, 'unknown')
-                        for license in self.upt_pkg.licenses]) or 'unknown'
+        licenses = []
+        for license in self.upt_pkg.licenses:
+            port_license = spdx2macports.get(license.spdx_identifier,
+                                             'unknown')
+            if license.spdx_identifier != 'unknown'and port_license == 'unknown': # noqa
```
Steap: There should be a space before "and".

upt_macports/upt_macports.py
```diff
@@ -30,9 +30,15 @@ def licenses(self):
...
+                self.logger.error(f'spdx2macports license conversion failed for {license.spdx_identifier}') # noqa
```
Steap: Why the "# noqa" on these lines?

Also, is this really an error? Shouldn't it rather be a warning?

upt_macports/upt_macports.py
```diff
@@ -30,9 +30,15 @@ def licenses(self):
...
+            else:
+                licenses.append(port_license)
```
Steap, line 46: Should we log this? 

> self.logging.info (f'Found license {port_license)')

upt_macports/upt_macports.py
Korusuke: for the second line, it would be necessary I guess, else the indentation would look very weird.

I was confused about error vs warning but went with error as it is something that should be fixed and if we go with a warning, people would tend to ignore it.

Steap: Error kind of gives the idea that something literally failed (even though we have ".critical()" also). What are the users supposed to do about this warning?

Isn't this more readable? 

```python
        licenses = []
        for license in self.upt_pkg.licenses:
            try:
                if license.spdx_identifier == 'unknown':
                    port_license = 'unknown'
                else:
                    port_license = spdx2macports.get[license.spdx_identifier]
                licenses.append(port_license)
            except KeyError:
                err = f'Unknown license {license.spdx_identifier}'
                self.logger.error(err)
        return ' '.join(licenses) or 'unknown'

```

reneeotten:
this looks good to me from a code perspective. I haven't tested it yet on some real-life examples (will do that tomorrow or so), but I assume @Korusuke that is tested and works as intended?

Korusuke:
Yes, it works as intended.

reneeotten:
okay, not necessarily related to the actual logging of the licenses as that seems to work as intended, but something I noticed this when trying out this feature. 

Let's take ```lmfit``` as an example, we have both a ```license``` field in the ```setup.py``` but also have a correct Trove classifier added. Admittedly, that's probably not how it should be done and I will change that with an upcoming PR.... Another one is ```chemex```, where again the Trove classifier for "BSD license" is correct, and the ```license``` field is used to specify that it is ```3-Clause BSD```.

In my opinion, upt should be able to get the license correct in these cases. I am fine with it ignoring the ```license``` field if that's deemed unreliable, but then it should just take the Trove classifier is present (which are in the ```json``` data for both packages). I haven't looked in detail so I might be missing something here, but would it make sense to try and adapt the logic somewhere so that this will work (@Steap)?

Steap:
Yeah, upt-pypi could probably do things better, but there will always be case where it's too hard to guess.

Regarding this patch, it lacks unit tests.

reneeotten:
Sure, but in the examples I mentioned there is no guessing needed since the correct license classifiers are present. I will take a closer look at upt-pypi, see why it behaves that way and suggest a change to improve it. 

Oh, I see... you have added ```'License :: OSI Approved :: BSD License'``` to "not_actual_licenses" since it doesn't specify the version. I agree that while that's perhaps not fully correct, it's anyway what we use for MacPorts as well... It's right now the way it works on PyPI and it would mean that all the pacakges with ```BSD``` as a license return "unknown" for MacPorts, that's not ideal.


Yes, it would indeed benefit from some tests - you’re right ;)

Korusuke:
I have added a "example" test.
I tried for few hours but couldn't get the logging-tests working :(
I guess it has something to do with `logging.getLogger()` but I am not sure.

Edit: The test has passed as there is no real comparison taking place

Korusuke:
I am stupid, should have just looked into existing upt code instead of trying to search online. The style in upt logger tests worked flawlessly, although I would love to know why previous 2/3 methods were giving so many errors.

Steap:
Honestly, more than the logging itself, I'm interested in testing the return values of this function.

Korusuke:
> Honestly, more than the logging itself, I'm interested in testing the return values of this function.

That is already there so I thought you were talking about logging.

upt_macports/upt_macports.py
```diff
@@ -30,9 +30,25 @@ def licenses(self):
         filepath = pkg_resources.resource_filename(__name__, relpath)
         with open(filepath) as f:
             spdx2macports = json.loads(f.read())
+        licenses = []
```
reneeotten: please move this initialization of the empty list closer to where it's used, in other words just before the ```for license in self.upt_pkg.licenses``` (no need to define it before the ```if```-statement) and leave a newline here.

upt_macports/upt_macports.py
```diff
@@ -30,9 +30,25 @@ def licenses(self):
...
+                    port_license = 'unknown'
+                    self.logger.warning(f'upt failed to detect license')
+                else:
+                    port_license = spdx2macports[license.spdx_identifier]
+                    self.logger.info(f'Found license {port_license}')
+                licenses.append(port_license)
+            except KeyError:
+                err = f'Unknown license {license.spdx_identifier}'
+                self.logger.error(err)
+                self.logger.info('Please report the error at https://github.com/macports/upt-macports') # noqa
```
reneeotten, line 51: if we end up in this ```except``` block, it means that ```upt``` found a license but it's not converted yet into a license that MacPorts understands (i.e., is present in ```spdx2macports.json```). 

So if I read this correctly that keeps ```licenses = []```, as initialized and logging it as an error will not halt the script, right? So now if we do the return statement it will return ```''``` and put the line ```license ''``` in the Portfile? I don't think that's what we want (or at least what I prefer). What about making the error message a bit more specific and return that same message so that it will be added in the ```license``` field in the Portfile as well? Perhaps something like:
```err = f'MacPorts license unknown for {license.spdx_identifier}'```

reneeotten:
@Steap anything else left to do here?

Steap:
@reneeotten I think we are good!

Regarding the PyPI licenses, I'm trying out some changes, but to give you an idea of why it's difficult to guess the license, here are the 4589 different values used in the "license" field https://framabin.org/p/?f4a6b18cb018e3cc#+aVuHWohgiG8EKmsG/XfmhzOk68l45JtcYhhTAZpS1Q= .

reneeotten:
okay, I see the issue with guessing all the license correctly... but even just BSD/MIT/Apache would probably cover the majority already. I'll merge this PR now since it works as intended!

Steap:
@reneeotten I'm reworking the PyPI code that tries to guess the license. I'm testing it on the 5000 most downloaded packages.

Using only classifiers, I fail to identify ~3000 licenses. Using the license field, I only fail to find ~1800 packages. If I then try to read the LICENSE file from the package, I only fail to identify ~1200 packages.

So, it's getting better, but it 's still not going to work every time :) Expect improvements in the next version of upt-pypi.

PR #29, opened 2019-06-13
Add output function

Korusuke:
It's not yet totally complete as ruby PR is yet to be merged.

### DIFF
--- a/upt_macports/upt_macports.py
+++ b/upt_macports/upt_macports.py
@@ -4,6 +4,8 @@ import jinja2
 import pkg_resources
 import json
 import requests
+import os
+import sys
 
 
 class MacPortsPackage(object):
@@ -13,7 +15,42 @@ class MacPortsPackage(object):
     def create_package(self, upt_pkg, output):
         self.upt_pkg = upt_pkg
         self.logger.info(f'Hello, creating the package')
-        print(self._render_makefile_template())
+        if output is None:
+            print(self._render_makefile_template())
+        else:
+            self._create_output_directories(upt_pkg, output)
+            self._create_portfile()
+
+    def _create_output_directories(self, upt_pkg, output_dir):
+        """Creates the directory layout required"""
+        self.logger.info(f'Creating the directory structure in {output_dir}')
+        upt2macports = {
+            'cpan': 'perl',
+            'pypi': 'python',
+            'rubygems': 'ruby',
+        }
+        try:
+            port_category = upt2macports[self.upt_pkg.frontend]
+        except KeyError:
+            sys.exit(
+                f'Unknown port category for frontend {self.upt_pkg.frontend}')
+        folder_name = self._normalized_macports_folder(upt_pkg.name)
+        self.output_dir = os.path.join(
+            output_dir, port_category, folder_name)
+        try:
+            os.makedirs(self.output_dir, exist_ok=True)
+            self.logger.info(f'Created {self.output_dir}')
+        except PermissionError:
+            sys.exit(f'Cannot create {self.output_dir}: permission denied.')
+
+    def _create_portfile(self):
+        self.logger.info('Creating the Portfile')
+        try:
+            with open(os.path.join(self.output_dir, 'Portfile'), 'x',
+                      encoding='utf-8') as f:
+                f.write(self._render_makefile_template())
+        except FileExistsError:
+            sys.exit(f'Cannot create {self.output_dir}/Portfile: already exists.') # noqa
 
     def _render_makefile_template(self):
         env = jinja2.Environment(
@@ -105,6 +142,11 @@ class MacPortsPythonPackage(MacPortsPackage):
         if pypi_name != self.upt_pkg.name.lower():
             return pypi_name
 
+    @staticmethod
+    def _normalized_macports_folder(name):
+        name = name.lower()
+        return f'py-{name}'
+
 
 class MacPortsNpmPackage(MacPortsPackage):
     template = 'npm.Portfile'
@@ -129,6 +171,11 @@ class MacPortsPerlPackage(MacPortsPackage):
     @staticmethod
     def _normalized_macports_name(name):
         return name.replace('::', '-')
+
+    @staticmethod
+    def _normalized_macports_folder(name):
+        name = name.lower().replace('::', '-')
+        return f'p5-{name}'
 
     def _cpandir(self):
         pkg = self.upt_pkg
@@ -165,6 +212,11 @@ class MacPortsRubyPackage(MacPortsPackage):
     @staticmethod
     def _normalized_macports_name(name):
         return name
+
+    @staticmethod
+    def _normalized_macports_folder(name):
+        name = name.lower()
+        return f'rb-{name}'
 
 
 class MacPortsBackend(upt.Backend):
--- a/upt_macports/tests/test_macports_package.py
+++ b/upt_macports/tests/test_macports_package.py
@@ -73,6 +73,56 @@ class TestMacPortsPackageLicenses(unittest.TestCase):
         self.assertEqual(m_stderr.getvalue(), err)
 
 
+class TestDirectoryCreation(unittest.TestCase):
+    def setUp(self):
+        self.package = MacPortsPackage()
+        self.package.upt_pkg = upt.Package('foo', '42')
+        self.package.upt_pkg.frontend = 'pypi'
+
+    def test_create_directories_output_invalid_frontend(self):
+        self.package.upt_pkg.frontend = 'frontend'
+        with self.assertRaises(SystemExit):
+            self.package._create_output_directories(self.package.upt_pkg,
+                                                    '/ports/')
+
+    @mock.patch('os.makedirs')
+    @mock.patch.object(MacPortsPackage, '_normalized_macports_folder',
+                       create=True, return_value='py-foo')
+    def test_create_directories_output(self, folder_name, m_mkdir):
+        self.package._create_output_directories(self.package.upt_pkg,
+                                                '/ports/')
+        m_mkdir.assert_called_with('/ports/python/py-foo', exist_ok=True)
+
+    @mock.patch('os.makedirs', side_effect=PermissionError)
+    @mock.patch.object(MacPortsPackage, '_normalized_macports_folder',
+                       create=True, return_value='py-foo')
+    def test_create_directories_permission_error(self, folder_name, m_mkdir):
+        with self.assertRaises(SystemExit):
+            self.package._create_output_directories(self.package.upt_pkg,
+                                                    '/ports/')
+
+
+class TestFileCreation(unittest.TestCase):
+    def setUp(self):
+        self.package = MacPortsPackage()
+        self.package.output_dir = '/outdir'
+        self.package.upt_pkg = upt.Package('foo', '42')
+
+    @mock.patch('builtins.open', new_callable=mock.mock_open)
+    def test_portfile_creation(self, m_open):
+        fn = 'upt_macports.upt_macports.MacPortsPackage._render_makefile_template' # noqa
+        with mock.patch(fn, return_value='Portfile content'):
+            self.package._create_portfile()
+            m_open.assert_called_once_with('/outdir/Portfile', 'x',
+                                           encoding='utf-8')
+            m_open().write.assert_called_once_with('Portfile content')
+
+    @mock.patch('builtins.open', side_effect=FileExistsError)
+    def test_portfile_file_exists(self, m_open):
+        with self.assertRaises(SystemExit):
+            self.package._create_portfile()
+
+
 class TestMacPortsPackageArchiveType(unittest.TestCase):
     def setUp(self):
         self.package = MacPortsPackage()


upt_macports/upt_macports.py
```diff
@@ -82,6 +118,10 @@ def _pkgname(self):
     def _normalized_macports_name(name):
         return f'py-{name}'
 
+    @staticmethod
+    def _normalized_macports_folder(name):
+        name = name.lower()
+        return f'py-{name}'
```
reneeotten, line 148: this is not correct; the directory name should be the same as the ```field``` name in the Portfile. There we decided to follow the upstream name, including possible capitalization

upt_macports/upt_macports.py
```diff
@@ -13,7 +15,41 @@ def __init__(self):
...
+            'rubygems': 'ruby',
+        }
+        try:
+            port_category = upt2macports.get(self.upt_pkg.frontend)
+        except KeyError:
+            sys.exit(
+                f'Unknown port category for frontend {self.upt_pkg.frontend}')
+        folder_name = self._normalized_macports_folder(upt_pkg.name)
+        self.output_dir = os.path.join(
+            output_dir, 'ports', port_category, folder_name)
```
reneeotten: I am not sure about this... why is there a hard-coded 'ports' in the ```output_dir```? I assume that most developers will have a [local copy of the ports-tree](https://guide.macports.org/chunked/development.local-repositories.html); for example I have mine in ```~/Software/macports-ports```. In that case I should be able to use that as my "ouput" directory and that doesn't work right now.

upt_macports/upt_macports.py
```diff
@@ -13,7 +15,41 @@ def __init__(self):
...
+        try:
+            os.makedirs(os.path.join(self.output_dir))
+        except PermissionError:
+            sys.exit(f'Cannot create {self.output_dir}: permission denied.')
+        except FileExistsError:
+            sys.exit(f'Cannot create {self.output_dir}: already exists.')
+
+    def _create_portfile(self):
+        self.logger.info(f'Creating the Portfile')
```
reneeotten: no need for a f-string here if you don't substitute anything

upt_macports/upt_macports.py
mojca: I agree that those two should match (and I would use a function to construct the name here), but I still don't like the idea of native capitalisation, in particular because it may differ on github and pypi.

I'm not sure if devs would have the same opinion after seeing a subset of PR camelcasing names.

upt_macports/upt_macports.py
mojca: I believe one of the tools used the `dports` prefix, but I don't like that either.

upt_macports/upt_macports.py
reneeotten: Using a function here would indeed make sense. I don't understand your comment about possible differences between GitHub and PyPI though, can you elaborate on that? From what I have seen pretty much anywhere people use whatever name "upstream" has decided to call the package (including capitalization).

The only people who responded on the mailing list to this topic were Rainer, Josh, and Ryan - and from what I gathered from there is that "following upstream would be fine". So either other devs don't care or have no opinion... But again, I really don't have a strong preference and it seems there are currently only 12 ports that do have capitalization, so I am fine with going to only lower case as well. That would be more consistent with the Perl packages - we just need to make an executive decision now and move on ;)

Korusuke: I just did a quick analysis of pypi package names and in that 20108/183646 have some type of capitalization which is around 10%.

Seeing this and the fact that I have never cared about the capitalization while installing packages via pip, I think we should follow lower case and this will also be consistent with perl packages as point be @reneeotten

mojca: > I just did a quick analysis of pypi package names and in that 20108/183646 have some type of capitalization which is around 10%.

This might be something worth mentioning in the discussion about the issue.
 
> Seeing this and the fact that I have never cared about the capitalization while installing packages via pip, I think we should follow lower case and this will also be consistent with perl packages as point be @reneeotten

That makes at least three of us :)

What I meant with potentially different capitalisation was also mention in my response to your query. If we had a package `py-macports`: how would you capitalise it? Note that all of our repository names are lowercase, yet our official name is CamelCase (MacPorts).

We seem to have just 12 ports starting with uppercase. I checked a random one, `py-SDL2`, and that one contains
```tcl
github.setup        marcusva py-sdl2 ...
```
If this wasn't a python module, we tend to use the exact same name as the github repository, which in this case is `py-sdl2` (lowercase), which could be a hint that we might be using "wrong" capitalisation, and if you look at `setup.py` you'll see:
```python
       "url": "https://github.com/marcusva/py-sdl2",
        "download_url": "https://pypi.python.org/pypi/PySDL2",
```
Now: which capitalisation is "correct"? If you check the contents of the package, it's mostly lowercase. Also, it's
```python
from sdl2 import ...
```
so lowercase when used as a python module name.

Another confusion of course is whether this should be `py-pysdl2` or `py-sdl2`, but that's yet a different topic.

If MacPorts is using a roughly representative sample of `pypi`, this means that we would need to rename roughly 150 ports, and also fix all ports which have those ports declared as dependencies.

The following comment is taken out of context, but I would prefer if we used something like
```tcl
pypi.setup PySDL2 0.9.6
```
or
```tcl
pypi.setup upt 0.7
```
which would automatically set name, version, python.rootname, master_sites, distname etc. Similar to how little the perl modules need.

Korusuke:
When trying to output to a file as of now it is throwing an error if the file already exists and the program stops.
I added recursive functionality to my local copy and when experimenting with that this was the biggest issue, like if we try packaging `upt-cpan` then it will add `requests` too and `requests` is also a dependency for `upt-pypi` so when trying to recursively package that it throws an error and stop there.

This could be a potential issue.
We can either just log the error and continue executing or we can create `.new` file in case the file already exists.

Also, another thing not related to this PR that I wanted to confirm, while recursive packaging do we need to check the MacPorts tree that whether the ports already exist or not ?

upt_macports/upt_macports.py
reneeotten: okay, let's then move to lowercase names for Python ports and let's stick with this decision at least during the remainder of the GSoC ;) 

@Korusuke please open another PR to make this change and keep in mind the discussion in e-mails we had about the need to set ```python.rootname``` to the upstream name if it differs from the MacPorts package name. That's the easiest way to keep the remainder of the Python template the same. 

@mojca I would consider suggestions about changing the Python PortGroup as outside of the scope of the ```upt``` project. Nevertheless, accommodating changes like that at a later stage in the template should be easy enough. 

About the question on how to name your examples ("PySDL2"): since we are having a PyPI frontend we will follow that name and add "py-" in front of it, so it will become "py-pysdl2". If upstream cannot decide on a consistent naming scheme themselves, I don't consider that our problem ;)

reneeotten:
> When trying to output to a file as of now it is throwing an error if the file already exists and the program stops.

Yes, and that's fine for this PR where the point is to write the output of one package to a file. Indeed it will need some more considerations later in the context of both recursive addition and updating. That discussion does not really belong in this PR though, so please open a new issue for this or start a discussion in the PR where it applies to.

> I added recursive functionality to my local copy and when experimenting with that this was the biggest issue, like if we try packaging `upt-cpan` then it will add `requests` too and `requests` is also a dependency for `upt-pypi` so when trying to recursively package that it throws an error and stop there.
> 
> This could be a potential issue. We can either just log the error and continue executing or we can create `.new` file in case the file already exists.
> 
> Also, another thing not related to this PR that I wanted to confirm, while recursive packaging do we need to check the MacPorts tree that whether the ports already exist or not ?

Short answer: yes, I would prefer to check if a package is already present (and whether the available version can fulfill the requirement). If so, I think it can be ignored from the perspective of "recursive packaging" - if you want to be pro-active one could check if it's actually the most recent version that is present and log a message if that's not the case.

Steap:
> When trying to output to a file as of now it is throwing an error if the file already exists and the program stops.
> I added recursive functionality to my local copy and when experimenting with that this was the biggest issue, like if we try packaging `upt-cpan` then it will add `requests` too and `requests` is also a dependency for `upt-pypi` so when trying to recursively package that it throws an error and stop there.
> 

I'm pretty sure we will want the backend to tell upt whether the package already exists.

upt_macports/upt_macports.py
Korusuke: > okay, let's then move to lowercase names for Python ports and let's stick with this decision at least during the remainder of the GSoC ;)
> 
> @Korusuke please open another PR to make this change and keep in mind the discussion in e-mails we had about the need to set `python.rootname` to the upstream name if it differs from the MacPorts package name. That's the easiest way to keep the remainder of the Python template the same.

I don't get why would we need to set `python.rootname` as from what I understand it's only used if the package name is different from MacPorts name except for capitalization but that won't be the case for upt ever

reneeotten: Please take a look at, for example, ```py-sphinx```. This is all lowercase in MacPorts, but upstream/PyPI uses ```Sphinx```. In this case for MacPorts the name is ```py-sphinx```, and ```python.rootname``` is set to ```sphinx``` by the Python PortGroup. Now, if you look in our python.Portfile template you will notice that this variable is used for things like ```master_sites``` and ```distname```, which are now not correct anymore because of the capitalization. At least I am pretty sure that's the case, but please verify this yourself by editing that Portfile and replacing all occurrences of "Sphinx" with "sphinx" in the master_sites and distname as that's the result you would get with upt.

In cases like this you can either set ```python.rootname``` to mach upstream, in the above example to ```Sphinx``` and leave the rest of the template untouched. Alternatively, you would need to change the ```distname``` and ```master_sites``` as is currently done in the ```py-sphinx``` Portfile. I think the first option is easier.

Korusuke: Okay Thanks for the explanation, I thought capitalization won't matter as it doesn't matter with urls but forgot that while downloading its actually a directory structure.

Added a PR as per this but while trying it out it somewhat works....more description in the PR.

jmroot: There are a lot of existing ports not using python.rootname simply because it's a relatively recent addition.

reneeotten:
> I'm pretty sure we will want the backend to tell upt whether the package already exists.

I can imagine that a developer wants to just generate a new Portfile (e.g., for comparison) and in that case ```upt``` should just print it to the screen even if the port already exists. So when using the "output" feature I suppose that the backend could check whether the port exists before attempting to package it again and then fail at the output stage, but for one port it actually doesn't take any time - so I would be fine with just checking it at then end when trying to write the file. 

I would prefer to change what happens during the ```except FileExistsError:``` to not do a ```sys.exit()``` but just log that the directory already exists and revert to printing the Portfile to stdout.

Nevertheless, in the long run for "recursive packaging" and the "update feature", we'll probably need to handle the situation where a port already exists differently and, for example, do an "update" in that case even if the end-user did "package" to start with.

@Korusuke @Steap what do you think, would this be a good (temporary) solution?

reneeotten:
ping @Steap @Korusuke see my comment above, what do you think about that suggestion?

Steap:
I think most of this is up to the various communities, and how they usually work. 

> Nevertheless, in the long run for "recursive packaging" and the "update feature", we'll probably need to handle the situation where a port already exists differently and, for example, do an "update" in that case even if the end-user did "package" to start with.

I'm not a big fan of this. The update feature is probably going to be kind of experimental for a while, so I would rather not update anything when the "--update" is not given.

reneeotten:
> I think most of this is up to the various communities, and how they usually work.
> 
> > Nevertheless, in the long run for "recursive packaging" and the "update feature", we'll probably need to handle the situation where a port already exists differently and, for example, do an "update" in that case even if the end-user did "package" to start with.
> 
> I'm not a big fan of this. The update feature is probably going to be kind of experimental for a while, so I would rather not update anything when the "--update" is not given.

@Steap, sorry for not being clear - I completely agree with your statement. What I meant to say is that once we have added the "--update" options we can of course not stop (or only print the Portfile to stdout) anymore when a port already exists. 

In other words, this PR looks good to me after the change below and we can merge this; we will need to revisit this when implementing the "update" feature.

>I would prefer to change what happens during the except FileExistsError: to not do a sys.exit() but just log that the directory already exists and revert to printing the Portfile to stdout.

mojca:
Just a random thought (not too relevant for this PR).

I'm currently spending waaaaay too much time just trying to get one single port packaged (and trying to be a good citizen by adding `test.run yes` to the ports when I can).

Complaining that upt gets dependencies wrong in most of the cases is a subject for another evening, I assume we could leverage this project "create a CI system for pypi packages" :), but what would definitely help would be upt telling me that "hey, these dependencies are outdated" (including recursive test dependencies). Sure, there could be an option to update ports at some point in the future, but already the info that something needs to be updated would be helpful. It would also help if we could get a warning when a particular version of dependency is not compatible (usually too old) or a dependent port requires an older version. (Getting a warning when you update a port to a version that's no longer compatible with a dependent port could be nice, but probably require a disproportionate amount of effort.)

Sadly `port livecheck rdepof:py37-foo` doesn't work because livechecks are disabled for subports and main port doesn't have the dependencies declared. But maybe there's a way to fix that (something to be discussed by the community).

reneeotten:
> In other words, this PR looks good to me after the change below and we can merge this; we will need to revisit this when implementing the "update" feature.
> 
> > I would prefer to change what happens during the except FileExistsError: to not do a sys.exit() but just log that the directory already exists and revert to printing the Portfile to stdout.

I am clearly not really thinking before typing.... I think it's fine like it is. One can, of course, just skip the "--output" and it will just print things to stdout ;)

reneeotten:
> Just a random thought (not too relevant for this PR).
> 
> I'm currently spending waaaaay too much time just trying to get one single port packaged (and trying to be a good citizen by adding `test.run yes` to the ports when I can).

Yes, it does take time to package something properly and be a good citizen.... but, without upt it would take probably even more time ;) 

> Complaining that upt gets dependencies wrong in most of the cases is a subject for another evening, I assume we could leverage this project "create a CI system for pypi packages" :), 

Well, unfortunately that's mainly upstream and/or PyPI their fault and not upt - unless you're saying that pypi2port produces better results; in that case we should certainly try and improve the current status of upt-macports.

> but what would definitely help would be upt telling me that "hey, these dependencies are outdated" (including recursive test dependencies). Sure, there could be an option to update ports at some point in the future, but already the info that something needs to be updated would be helpful. It would also help if we could get a warning when a particular version of dependency is not compatible (usually too old) or a dependent port requires an older version. (Getting a warning when you update a port to a version that's no longer compatible with a dependent port could be nice, but probably require a disproportionate amount of effort.)

Much of what you're discussing here is part of the "recursive packaging" feature that is being discussed/implemented in #32 and #35. I appears that in general the specifiers for versioning for all frontends isn't that robust (yet), but I think for PyPI we should be able to most of what you're suggesting.

Korusuke:
> I am clearly not really thinking before typing.... I think it's fine like it is. One can, of course, just skip the "--output" and it will just print things to stdout ;)

So I guess this PR is ready to be merged or should we change the `sys.exit` to just logging as this will be needed in recursive packaging?

reneeotten:
okay, I am going to merge this by tomorrow morning (so in let's say 15 hours from now) unless I hear otherwise! Confirming that you (@Steap) agree with it (or just merging it yourself) would shorten this timeline ;)

Steap:
My main issue is that there are no tests here. Creating directories, making sure we do not overwrite anything, etc.: these actions deserve some testing.

reneeotten:
> My main issue is that there are no tests here. Creating directories, making sure we do not overwrite anything, etc.: these actions deserve some testing.

I didn't look at that, but it's a good point - the coverage would indeed go down from 84 to 71%. So the output function itself looks good when we're adding a single port and to get this finished @Korusuke please add tests. Also, please rebase with ```master``` and resolve possible conflicts.

After that we can move on to the recursive part of the packaging, and then the output function might need some additional tweaks but that should not stop us from finishing up this PR now.

reneeotten:
I do not have much experience with ```mock```, but reading a bit through the documentation and your tests it seems that it could work like this. You're now "mimicking/mocking" situations pretending that you would get a PermissionError or FileExistsError. I would probably have made a temporary directory and made sure that you generate the correct directory/Portfile and that the proper errors are raised where you want them to (right now you only test whether a SystemExit happens), but in the end it might be equivalent. 

One situation that isn't handled now is the following:
- directory already exists (for some reason already made by the end-user), but there is no file in there.

Currently, it will throw a ```FileExistsError``` which actually just checks if the directory exists; this check could perhaps be expanded?

@Steap do you have (more) experience with ```mock``` and are these tests sufficient?

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "python.pythonPath": "/usr/bin/python3.7"
+}
```
Steap: Please check "git status" before committing, and refrain from using "git commit -a" too much :)

lib64
```diff
@@ -0,0 +1 @@
+lib
```
Steap: Ditto.

pip-selfcheck.json
```diff
@@ -0,0 +1 @@
+{"last_check":"2019-06-13T03:14:34Z","pypi_version":"19.1.1"}
```
Steap: Ditto.

pyvenv.cfg
```diff
@@ -0,0 +1,3 @@
+home = /home/karan/anaconda3/bin
+include-system-site-packages = false
+version = 3.6.5
```
Steap: Ditto.

Steap:
@reneeotten os.makedirs takes an optional argument: exist_ok. If set to True, os.makedirs will not fail if a directory already exists. By default, it is set to False.

Other than that, the tests look ok.

upt_macports/upt_macports.py
```diff
@@ -13,7 +15,41 @@ def __init__(self):
     def create_package(self, upt_pkg, output):
         self.upt_pkg = upt_pkg
         self.logger.info(f'Hello, creating the package')
-        print(self._render_makefile_template())
+        if output is None:
+            print(self._render_makefile_template())
+        else:
+            self._create_output_directories(upt_pkg, output)
+            self.logger.info('Directory structure created')
```
Steap: This does not give enough info to the user. We should tell them "Created /tmp/foo/bar".

Korusuke:
> @reneeotten os.makedirs takes an optional argument: exist_ok. If set to True, os.makedirs will not fail if a directory already exists. By default, it is set to False.
> 
> Other than that, the tests look ok.

Just to make sure, here it means that if the folder structure already exists then don't throw an error but then if file exists then do throw an error, right?

reneeotten:
> > @reneeotten os.makedirs takes an optional argument: exist_ok. If set to True, os.makedirs will not fail if a directory already exists. By default, it is set to False.
> > Other than that, the tests look ok.
> 
> Just to make sure, here it means that if the folder structure already exists then don't throw an error but then if file exists then do throw an error, right?

@Steap @Korusuke yes I am aware of the ```exist_ok``` argument for ```os.makedirs``` and that would probably be a good idea to use. One could catch then only the ```PermissionError``` and do an additional check if a ```Portfile``` exists in the ```ouput_dir```. 

The most important check is whether the Portfile is present and if so to make sure we don't overwrite anything. I would be in favor of continuing if for some reason the end-user already made the directory, but there isn't a Portfile yet.

upt_macports/upt_macports.py
```diff
@@ -13,7 +15,42 @@ def __init__(self):
...
+        try:
+            port_category = upt2macports[self.upt_pkg.frontend]
+        except KeyError:
+            sys.exit(
+                f'Unknown port category for frontend {self.upt_pkg.frontend}')
+        folder_name = self._normalized_macports_folder(upt_pkg.name)
+        self.output_dir = os.path.join(
+            output_dir, port_category, folder_name)
+        try:
+            os.makedirs(os.path.join(self.output_dir), exist_ok=True)
```
Steap: os.path.join with a single argument seems a bit useless here, isn't it?

reneeotten: true ;) Besides that, I think this is ready! I'll do some final testing tomorrow morning, but if nothing pops up there, any concerns with me merging this PR?

reneeotten:
okay, in it goes!

PR #33, opened 2019-06-17
Change python naming convention

Korusuke:
python.rootname is set if there's capitalization but this check is performed on the name supplied by the user and not on the actual name.

For example for `WTForms` and `wtforms` file is generated but one with the `rootname` field and one without which should not be usually there.
For solving this `upt-pypi` should give the actual package name instead of the name supplied by the user.

### DIFF
--- a/upt_macports/templates/python.Portfile
+++ b/upt_macports/templates/python.Portfile
@@ -5,6 +5,9 @@ PortGroup           python 1.0
 {% endblock %}
 {% block nameversion %}
 name                {{ pkg._pkgname() }}
+{% if pkg._python_root_name() %}
+python.rootname     {{ pkg._python_root_name() }}
+{% endif %}
 version             {{ pkg.upt_pkg.version }}
 revision            0
 {% endblock %}
--- a/upt_macports/tests/test_python_package.py
+++ b/upt_macports/tests/test_python_package.py
@@ -11,7 +11,7 @@ class TestMacPortsPythonPackage(unittest.TestCase):
         self.package.upt_pkg = upt.Package('test-pkg', '13.37')
 
     def test_pkgname(self):
-        expected = ['py-foo', 'py-py-foo', 'py-pyfoo', 'py-pyFoo']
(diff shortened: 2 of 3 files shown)


reneeotten:
ah, I see - so the package name depends on whether the user provides the port with/without capitalization; the rest of the Portfile is the same because the capitalization apparently doesn't matter for the PyPI url.

That is indeed a problem, but should be easy to solve as the actual PyPI name is present as ```self.json['info']['name']``` it just not returned in the ```parse``` function of upt-pypi. @Steap what do you think?

Steap:
OK, I'll include that in the next upt-pypi version.

reneeotten:
sounds good @Steap! I would really like to get the templates including this naming business finalized as soon as possible, so depending on how long that's going to take, it would be easy to add a workaround in the backend for this right now. You can get the *exact* PyPI name by using
```
pypi_name = self.upt_pkg.get_archive().filename.split('-'+self.upt_pkg.version)[0])
```

reneeotten:
@Korusuke @Steap so what's your opinion on this?

Korusuke:
I guess even if name source is changed in `upt-pypi`, for backends it will be the same variable so we can merge this PR without waiting for that currently.

reneeotten:
> I guess even if name source is changed in `upt-pypi`, for backends it will be the same variable so we can merge this PR without waiting for that currently.

Right, but in the PR right now the behavior depends on what the end-user gives as a package name (e.g., Spinxh versus sphinx) and can result in a non-working Portfile - that should not be the case. So, we can either what for the functionality to be added to ```upt-pypi``` or, as I would prefer, you make the change I suggested, we merge this PR and once ```upt-pypi``` has gained the "real_name" functionality we change to that.

reneeotten:
thanks @Korusuke this works fine now so I am going to merge it! Once ```upt-pypi``` returns the "real" PyPI package name we can use that instead of extracting it from the archive filename.

PR #34, opened 2019-06-18
Add proper content type for our README.

Steap:
This makes sure the README is displayed correctly on PyPI.


reneeotten:
that looks like a rather uncontroversial change and is fine with me. I suppose for now we don't need to publish it to PyPI yet, but we can do so later. Before doing that we might need to update some more fields.

Issue #32, opened 2019-06-17
Recursive Packaging


Korusuke:
As already done in the [current PR](https://framagit.org/upt/upt/merge_requests/1), now `args.packages` is a list which upt iterates over and generates all the ports.

So using this list we can simply add a recursive feature like once the frontend returns the list of required packages.
Add them to the list if not already in the list. But in this, there is no check or control by backend which can be added if instead of upt adding to list, the backend does that.

Now the question is in the backend should we "regenerate" ports if already present in MacPorts tree and second, currently if we output to file the program stop on encountering an existing file which would be a problem if we try recursively packaging ports having the same dependency not committed to MacPorts tree.
So should we change this to just logging the error and continue?

Steap:
If I'm not mistaken, your upt PR allows users to ask upt to package more than one piece of software. I don't think it is exactly "recursive" packaging.

Steap:
> So using this list we can simply add a recursive feature like once the frontend returns the list of required packages.
> Add them to the list if not already in the list. But in this, there is no check or control by backend which can be added if instead of upt adding to list, the backend does that.

How do we handle loops in our dependency graph? Especially since we are building it as we go?

Should we do a depth-first exploration or a breadth-first exploration?

> Now the question is in the backend should we "regenerate" ports if already present in MacPorts tree 

Probably not.

> currently if we output to file the program stop on encountering an existing file which would be a problem if we try recursively packaging ports having the same dependency not committed to MacPorts tree.

Well, we're supposed to remember what we've packaged so far, aren't we?

Korusuke:
> How do we handle loops in our dependency graph? Especially since we are building it as we go?

Since we are building as we go we can simply check while adding the dependency if its already in the list or not.

> Should we do a depth-first exploration or a breadth-first exploration?

Currently what I tried is breadth-first.
Here is the code that I am using currently (can be cleaned and optimized further):
```
for pkg in args.package:
            try:
                upt_pkg = None
                upt.log.logger_set_formatter(logger, 'Frontend')
                upt_pkg = frontend.parse(pkg)
                upt_pkg.frontend = args.frontend
                upt.log.logger_set_formatter(logger, 'Backend')
                backend.create_package(upt_pkg, output=args.output)
                deps = []
                for i in upt_pkg.requirements.get('run',[]):
                    deps.append(i.name)
                for i in upt_pkg.requirements.get('test', []):
                    deps.append(i.name)
[...]

PR #35, opened 2019-06-19
Recursive Packaging

Korusuke:
This is by no means complete, but wanted to share and get input on whether is it right and what else can we do.

For this to work changes required on `upt` side will be this:
```
packages = list(args.package)
i = 0
while i < len(packages):
      try:
           upt_pkg = None
           upt.log.logger_set_formatter(logger, 'Frontend')
           upt_pkg = frontend.parse(packages[i])
           upt_pkg.frontend = args.frontend
           upt.log.logger_set_formatter(logger, 'Backend')
           upt_pkg.packages = list(packages)
           upt_pkg = backend.create_package(upt_pkg, output=args.output)
           packages = list(upt_pkg.packages)
           i += 1
```

So currently as you can see, the code doesn't check MacPorts Tree and neither ask the user.
I tested this out for `upt-cpan` and it generated this list:

```
['upt-cpan', 'requests', 'upt', 'requests-mock', 'chardet', 'idna', 
'urllib3', 'certifi', 'spdx-lookup', 'six', 'fixtures', 'mock', 
'purl', 'pytest', 'sphinx', 'testrepository', 'testtools', 'hypothesis',
 'pbr', 'pytest-cov', 'py', 'packaging', 'attrs', 'atomicwrites', 'pluggy', 
 'importlib-metadata', 'wcwidth', 'argcomplete', 'nose', 'sphinxcontrib-applehelp', 
 'sphinxcontrib-devhelp', 'sphinxcontrib-jsmath', 'sphinxcontrib-htmlhelp', 
 'sphinxcontrib-serializinghtml', 'sphinxcontrib-qthelp', 'Jinja2', 'Pygments', 
 'docutils', 'snowballstemmer', 'babel', 'alabaster', 'imagesize', 'setuptools', 
 'html5lib', 'flake8', 'flake8-import-order', 'mypy', 'docutils-stubs', 'extras', 
 'python-mimeparse', 'traceback2', 'unittest2', 'coverage', 'fields', 'hunter', 
 'process-tests', 'virtualenv', 'pyparsing', 'pympler', 'zope.interface', 'zipp', 
 'pexpect', 'wheel', 'MarkupSafe', 'pytz', 'webencodings', 'entrypoints', 'pyflakes', 
 'pycodestyle', 'mccabe', 'typed-ast', 'mypy-extensions', 'linecache2', 'contextlib2', 
 'argparse', 'colorama', 'pytest-timeout', 'pytest-xdist', 'pytest-localserver', 
 'pypiserver', 'zope.event', 'pathlib2', 'ptyprocess', 'execnet', 'pytest-forked', 
 'filelock', 'zope.testrunner', 'apipkg', 'zope.exceptions', 'zope.testing']
```

Which as you can see is very big and my guess is not required for just `upt-cpan`.
If we just take the run dependencies then the output is similar to what pypi would install, which is:

`['upt-cpan', 'requests', 'upt', 'chardet', 'idna', 'urllib3', 'certifi', 'spdx-lookup']`

So if the way I am doing it right now is fine then next thing would be to add MacPorts tree check which can be done directly by "guessing" the dependency name
and a get request like for upt to this url
`https://raw.githubusercontent.com/macports/macports-ports/master/python/py-upt/Portfile`

Is there any better way to search github for package name in macports tree so that we don't have to "guess" the path and/or can we directly get the version of the package from MacPorts somehow.

and then comes version extraction and decision that whether a Portfile should be generated or not.

In this I am not sure if we should do this fully automatically, here we can keep user-input.

### DIFF
--- a/upt_macports/upt_macports.py
+++ b/upt_macports/upt_macports.py
@@ -5,6 +5,7 @@ import pkg_resources
 import json
 import requests
 import os
+import subprocess
 import sys
 
 
@@ -14,7 +15,7 @@ class MacPortsPackage(object):
 
     def create_package(self, upt_pkg, output):
         self.upt_pkg = upt_pkg
-        self.logger.info(f'Hello, creating the package')
+        self.logger.info(f'Creating MacPorts package for {self.upt_pkg.name}')
         if output is None:
             print(self._render_makefile_template())
         else:
@@ -227,20 +228,53 @@ class MacPortsRubyPackage(MacPortsPackage):
 
 
 class MacPortsBackend(upt.Backend):
+    def __init__(self):
+        self.logger = logging.getLogger('upt')
+
     name = 'macports'
+    pkg_classes = {
+        'pypi': MacPortsPythonPackage,
+        'cpan': MacPortsPerlPackage,
+        'rubygems': MacPortsRubyPackage,
+        'npm': MacPortsNpmPackage
+    }
 
     def create_package(self, upt_pkg, output=None):
-        pkg_classes = {
-            'pypi': MacPortsPythonPackage,
-            'cpan': MacPortsPerlPackage,
-            'rubygems': MacPortsRubyPackage,
-            'npm': MacPortsNpmPackage
-        }
-
         try:
-            pkg_cls = pkg_classes[upt_pkg.frontend]
+            self.frontend = upt_pkg.frontend
+            pkg_cls = self.pkg_classes[upt_pkg.frontend]
         except KeyError:
             raise upt.UnhandledFrontendError(self.name, upt_pkg.frontend)
-
         packager = pkg_cls()
         packager.create_package(upt_pkg, output)
+
+    def package_versions(self, name):
+        try:
+            port_name = self.pkg_classes[
+                self.frontend]._normalized_macports_folder(name)
+        except KeyError:
+            raise upt.UnhandledFrontendError(self.name, self.upt_pkg.frontend)
+
+        self.logger.info(f'Checking MacPorts tree for port {port_name}')
+        cmd = f'port info --version {port_name}'
+        port = subprocess.getoutput(cmd)
+        if port.startswith('Error'):
+            self.logger.info(f'{port_name} not found in MacPorts tree')
+            return []
+        elif port.startswith('version'):
+            curr_ver = port.split()[1]
+            self.logger.info(
+                f'Current MacPorts Version for {port_name} is {curr_ver}')
+            return [curr_ver]
+        elif port.startswith('Warning'):
+            self.logger.warning(
+                'port definitions are more than two weeks old, '
+                'consider updating them by running \'port selfupdate\'.')
+            curr_ver = port.split('version: ')[1]
+            self.logger.info(
+                f'Current MacPorts Version for {port_name} is {curr_ver}')
+            return [curr_ver]
+        else:
+            sys.exit(f'The command "{cmd}" failed. '
+                     'Please make sure you have MacPorts installed '
+                     'and/or your PATH is set-up correctly.')
--- a/upt_macports/tests/test_macports_backend.py
+++ b/upt_macports/tests/test_macports_backend.py
@@ -1,4 +1,5 @@
 import unittest
+from unittest import mock
 import upt
 from upt_macports.upt_macports import MacPortsBackend
 
@@ -12,6 +13,40 @@ class TestMacPortsBackend(unittest.TestCase):
         upt_pkg.frontend = 'invalid frontend'
         with self.assertRaises(upt.UnhandledFrontendError):
             self.macports_backend.create_package(upt_pkg)
+
+
+class TestMacPortsPackageExist(unittest.TestCase):
+    def setUp(self):
+        self.macports_backend = MacPortsBackend()
+        self.macports_backend.upt_pkg = upt.Package('foo', '42')
+        self.macports_backend.frontend = 'pypi'
+
+    @mock.patch('subprocess.getoutput')
+    def test_port_found(self, mock_sub):
+        expected = ['0.123']
+        mock_sub.return_value = 'version: 0.123'
+        self.assertEqual(
+            self.macports_backend.package_versions('foo'), expected)
+
+    @mock.patch('subprocess.getoutput')
+    def test_port_not_found(self, mock_sub):
+        expected = []
+        mock_sub.return_value = 'Error: fake-error'
+        self.assertEqual(
+            self.macports_backend.package_versions('foo'), expected)
+
+    @mock.patch('subprocess.getoutput')
+    def test_port_outdated(self, mock_sub):
+        expected = ['0.123']
+        mock_sub.return_value = 'Warning: fake-warning \nversion: 0.123'
+        self.assertEqual(
+            self.macports_backend.package_versions('foo'), expected)
+
+    @mock.patch('subprocess.getoutput')
+    def test_port_error(self, mock_sub):
+        mock_sub.return_value = 'bash: port: command not found'
+        with self.assertRaises(SystemExit):
+            self.macports_backend.package_versions('foo')
 
 
 if __name__ == '__main__':


Steap:
> Which as you can see is very big and my guess is not required for just `upt-cpan`.

I don't understand how you came to that conclusion. If this is the list you computed, then it means that they are actually required.

> If we just take the run dependencies then the output is similar to what pypi would install, which is:
> 
> `['upt-cpan', 'requests', 'upt', 'chardet', 'idna', 'urllib3', 'certifi', 'spdx-lookup']`

But we probably do not want that. Why would we drop the test dependencies?

Korusuke:
> > Which as you can see is very big and my guess is not required for just `upt-cpan`.
> 
> I don't understand how you came to that conclusion. If this is the list you computed, then it means that they are actually required.
>
> > If we just take the run dependencies then the output is similar to what pypi would install, which is:
> > `['upt-cpan', 'requests', 'upt', 'chardet', 'idna', 'urllib3', 'certifi', 'spdx-lookup']`
> 
> But we probably do not want that. Why would we drop the test dependencies?

Sorry, I did not mean that they are not required but even while installing via MacPorts currently not all of this ports are installed.
But by this, I am not implying that they are not required but maybe it works without them.

Other than this is it fine the way I am implementing the feature?

reneeotten:
we actually do install the same (or at least very similar list of Python packages for ```upt-cpan``` as well in MacPorts. The difference you noticed is due to the fact that we have ```upt``` depend on ```py-cpan``` to make sure we by default install the frontends, but that means we cannot declare a dependency on ```upt``` anymore for ```py-cpan``` since then we would have a circular dependency problem. 

So you should compare the Python ports we install when using ```upt``` (and ignore the other upt-frontends/backend): ```port rdeps upt | grep "py37-```
``` py37-spdx-lookup
    py37-setuptools
    py37-spdx
  py37-upt-cpan
    py37-requests
      py37-chardet
      py37-idna
      py37-urllib3
      py37-certifi
    py37-requests-mock
      py37-pbr
      py37-six
  py37-upt-macports
    py37-jinja2
      py37-markupsafe
        py37-pytest
          py37-setuptools_scm
          py37-py
          py37-attrs
            py37-hypothesis
              py37-coverage
            py37-zopeinterface
          py37-more-itertools
          py37-atomicwrites
          py37-packaging
            py37-parsing
          py37-pluggy
            py37-importlib-metadata
              py37-zipp
          py37-wcwidth
```

this looks already more complete, but I do agree that things are still seem missing. A big chunk of them seem related to test-dependencies which are indeed not always used in MacPorts.

Korusuke:
So as suggested by @reneeotten in #32 to use `port info <pkg-name>`.
I have added a preliminary version of that.

For version compatibility checking, I am not sure how would we do that as when we are adding a specific port to the generate list at that point we don't know the upstream version and only know the MacPorts version and the acceptable version for the parent pkg.

One possible solution which is a bit messy is to add the macports version as a property to the list and when we are to generate the package for that we compare the upstream version with the macports version and acceptable range, but this makes the code a bit hard to read.

As of now it is recursively packing only if the port is not available in the MacPorts tree without any version checking.


Another query I had is should we use `port info` as that make upt-macports os dependent.
All though a bit tricky but we can use github as suggested in the above comment.

Korusuke:
With this changes done, I would say the flow although a bit weird is good enough.
Only thing left is to add the constraint checking and then it would be ready to merge.
But it can be merged before that too as constraint checking may not be always reliable.

Current output for `upt package -f pypi -b macport upt --recursive` is:
```
[INFO    ] [Backend] Creating MacPorts package for upt
[INFO    ] [Backend] Checking MaPorts tree for port py-upt
[INFO    ] [Backend] Current MacPorts Version for py-upt is 0.7
[INFO    ] [Backend] Current Upstream version is 0.9
[INFO    ] [Backend] Creating Portfile as new version is available
[INFO    ] [Backend] Creating MacPorts package for spdx-lookup
[INFO    ] [Backend] Checking MaPorts tree for port py-spdx-lookup
[INFO    ] [Backend] Current MacPorts Version for py-spdx-lookup is 0.3.2
[INFO    ] [Backend] Current Upstream version is 0.3.2
[INFO    ] [Backend] Skipping spdx-lookup as MacPorts Portfile is already up-to-date
```

Any changes or review on the overall flow or the code itself ?

Steap:
Please show the upt code by sending a pull request to https://framagit.org/upt/upt .

> [INFO    ] [Backend] Checking MaPorts tree for port py-upt
> [INFO    ] [Backend] Current MacPorts Version for py-upt is 0.7
> [INFO    ] [Backend] Current Upstream version is 0.9

Looks like you're updating ports without the --update flag.

Steap:
Also, there is a lot of logic here that should be part of upt itself.

Korusuke:
Updated the existing pull request, will squash the commits once everything is finalized :)

I am aware that some of the logic could be transferred to upt itself but I was avoiding that as much as possible as any major change in upt would mean that all other backends would have to be updated too.
But please do point out anything that you would like to be moved to upt itself.

Korusuke:
> Please show the upt code by sending a pull request to https://framagit.org/upt/upt .
> 
> > [INFO    ] [Backend] Checking MaPorts tree for port py-upt
> > [INFO    ] [Backend] Current MacPorts Version for py-upt is 0.7
> > [INFO    ] [Backend] Current Upstream version is 0.9
> 
> Looks like you're updating ports without the --update flag.

No, I am not updating the ports.
It is just comparing the version and if new version is available than a new Portfile is created without using any data from the old Portfile.
Whereas in update feature a custom parser for MacPorts portfile will be required to update only the necessary fields.

Steap:
> I am aware that some of the logic could be transferred to upt itself but I was avoiding that as much as possible as any major change in upt would mean that all other backends would have to be updated too.
> But please do point out anything that you would like to be moved to upt itself.

Not necessarily. If a backend does not implement the "recursive" feature, it just means its users won't be able to use "--recursive", end of story.

The point of upt is to put everything complex in upt itself, and have easy-to-write modules. Think of the current state of the backends: a backend is just a few templates and some glue code that takes care of simple things (such as naming conventions) that are specific to this backend.

The only thing required on the backend side should be a "do you have this package already?" method. The complex code should be in upt itself and look something like (in pseudo-code):

```
function do_package(pkgname):
    pkg = frontend.parse(pkgname)
    backend.create_package(pkg)
    if !recursive:
        return
    for dep in pkg.all_dependencies():
        if backend.has_package(dep):
            continue
        do_package(dep)
```

Obviously we will need logging, error handling, etc. But first, we should get something like this to work. The initial implementation of has_package() could be "return True". Once this works, we can add special cases and various enhancements.

Korusuke:
Updated both the PR's, upt is currently failing on one test will fix it if the code changes looks good.
I am not so happy with how the upt code looks mainly because of no. of arguments in `pkg_it()`.

Now it almost follows the way @Steap suggested except that `has_package()` is replaced by `check_needed()` which also compares the version and can have more logic.

Yes we can have the version comparison logic in `upt` but I would say that different backends may decide differently based on versions and other parameters like in `ruby-gems` some backends wont consider the `configure` phase dependencies.

I tested the current PR with `upt` and `upt-cpan` as root packages and it works fine. 
Now it is Depth First Traversal unlike before.

reneeotten:
> Another query I had is should we use `port info` as that make upt-macports os dependent.
> All though a bit tricky but we can use github as suggested in the above comment.

You are right that the use of ```port``` to check whether a port is already present in the MacPorts tree and it's version information would require MacPorts to be installed. I am totally fine with that, you have recently installed MacPorts on Linux so it could still work there; I think that we can expect that anyone wanting to package something for MacPorts can actually run the ```port``` command.

upt_macports/upt_macports.py
```diff
@@ -176,6 +180,44 @@ def create_package(self, upt_pkg, output=None):
...
+
+    def pkg_exist(self, pkg_name, frontend):
+        prefix = {
+            'pypi': 'py-',
+            'rubygems': 'rb-',
+            'cpan': 'p5-',
+        }
+        port_name = prefix.get(self.upt_pkg.frontend, '') + self.upt_pkg.name
+        self.logger.info(f'Checking MaPorts tree for port {port_name}')
+        cmd = f'port info {port_name}'
```
reneeotten: Just as a comment: this will only check whether the port itself is present in the MacPorts tree, which does not necessarily mean that for example the ```pyXY-``` or ```rb-XY``` subport is present. But I think for now this is fine, I think down the road we'll add a ```port lint``` step which would check this then or we add logic for this at a later point - that should not delay the addition of this feature.

upt_macports/upt_macports.py
```diff
@@ -176,6 +180,44 @@ def create_package(self, upt_pkg, output=None):
...
+        port = os.popen(cmd).read()
+        self.logger.info(f'Current Upstream version is {self.upt_pkg.version}')
+        if port == '':
+            self.logger.info(f'{port_name} not found in MacPorts tree')
+            self.curr_ver = ''
+            return False
+        else:
+            self.curr_ver = port.split('@')[1].split()[0].split('_')[0]
```
reneeotten: if we are only going to extract version information, we can (above) run ```port info --version <portname>```, that will print either:
```
>>port info --version py-upt-pypi
version: 0.3
```

or 

```

>>port info --version portname
Error: Port portname not found
```

upt_macports/upt_macports.py
```diff
@@ -176,6 +180,44 @@ def create_package(self, upt_pkg, output=None):
...
         packager.create_package(upt_pkg, output)
+
+    def pkg_exist(self, pkg_name, frontend):
+        prefix = {
+            'pypi': 'py-',
+            'rubygems': 'rb-',
+            'cpan': 'p5-',
+        }
+        port_name = prefix.get(self.upt_pkg.frontend, '') + self.upt_pkg.name
+        self.logger.info(f'Checking MaPorts tree for port {port_name}')
```
reneeotten: ```MaPorts``` --> ```MacPorts```

upt_macports/upt_macports.py
```diff
@@ -176,6 +180,44 @@ def create_package(self, upt_pkg, output=None):
...
+                f'Current MacPorts Version for {port_name} is {self.curr_ver}')
+            return True
+
+    def check_needed(self, upt_pkg, constraint):
+        self.upt_pkg = upt_pkg
+        if self.pkg_exist(upt_pkg.name, upt_pkg.frontend):
+            curr_ver = self.curr_ver
+            upstream_ver = self.upt_pkg.version
+            # TODO: Add Decision Logic
+            self.logger.info(f'Constraint is {constraint}')
```
reneeotten: we probably should only log this if there is a constraint, or?

upt_macports/upt_macports.py
```diff
@@ -176,6 +180,44 @@ def create_package(self, upt_pkg, output=None):
...
+            if curr_ver < upstream_ver:
+                self.logger.info(
+                    'Creating Portfile as new version is available')
+                return True
```
reneeotten: this looks like the place where an "update" would be needed, right? Since we don't have code available to handle that, shouldn't we just log that there is a new upstream version but not return ```True``` as of yet?

upt_macports/upt_macports.py
```diff
@@ -176,6 +180,44 @@ def create_package(self, upt_pkg, output=None):
...
+    def pkg_exist(self, pkg_name, frontend):
+        prefix = {
+            'pypi': 'py-',
+            'rubygems': 'rb-',
+            'cpan': 'p5-',
+        }
+        port_name = prefix.get(self.upt_pkg.frontend, '') + self.upt_pkg.name
+        self.logger.info(f'Checking MaPorts tree for port {port_name}')
+        cmd = f'port info {port_name}'
+        port = os.popen(cmd).read()
```
reneeotten: I think we should be using ```subprocess``` here instead of ```os.popen``` (see [here](https://docs.python.org/3/library/subprocess.html#replacing-os-popen-os-popen2-os-popen3)). Currently, if a port isn't found it will print that output to the screen, we should probably just suppress that since we are already logging it.

reneeotten (commented):
a few comments inline

upt_macports/upt_macports.py
Korusuke: You are right that this is the scenario where we would update the Portfile but that is only if the `--update` is passed and as of now even if we don't have update working, it's like if the upstream has updated and the online MacPorts tree has older version, then we generate a new Portfile but if theres a local Portfile then it will anyways throw an error while writing.

So I think we should keep it as `True` as even after update is ready but for some reason user wants just the "pure" upt output then this will be operational.

upt_macports/upt_macports.py
Korusuke: Changed it to two different log statements.

reneeotten: that's fine as well, although if there is no version constraint that message is a bit useless - but if you see a use-case for it right now (for example for some debugging) am fine with keeping it as you have done now. We can reevaluate the amount of log messages we produce later when you add the colored output from your proposal.

Korusuke:
Tests added :)

Korusuke:
Anything else that needs to be modified in this or corresponding upt [PR](https://framagit.org/upt/upt/merge_requests/1) ?

upt_macports/upt_macports.py
```diff
@@ -176,6 +180,47 @@ def create_package(self, upt_pkg, output=None):
...
+            'pypi': 'py-',
+            'rubygems': 'rb-',
+            'cpan': 'p5-',
+        }
+        port_name = prefix.get(self.upt_pkg.frontend, '') + self.upt_pkg.name
+        self.logger.info(f'Checking MacPorts tree for port {port_name}')
+        cmd = f'port info --version {port_name}'
+        port = subprocess.getoutput(cmd)
+        self.logger.info(f'Current Upstream version is {self.upt_pkg.version}')
+        if port[:5] == 'Error':
```
reneeotten: I would personally prefer: ```if port.startswith('Error'):```

upt_macports/upt_macports.py
reneeotten: I don't think I agree with that ;) At least not if we don't add additional checks with respect to the constraints. If a package ```py-foo``` is happy with whatever version of ```py-bar```, then even if there is a newer version of ```py-bar``` available upstream we should not generate a new Portfile (or probably even update it unconditionally once that feature is available). There might be valid reason for keeping ```py-bar``` at a lower version because of requirements of other packages for example.

So I think we should check whether the available package would fulfill the ```constraints``` and if that's the case we just log "FYI: there is a newer version available upstream" but don't do anything else. If the version is too low then we should certainly log that and only in that case we could consider to return ```True```. But even there I would be in favor of not doing that right now as this would only work properly anyway once the "update" feature is in place. 

@Steap - opinions / insights?

upt_macports/upt_macports.py
Steap: I definitely agree.

upt_macports/upt_macports.py
Steap: I think you want to keep commits as small as possible, so you should not update without a valid reason. Think of a Python package that uses pytest, for instance. If you upgrade pytest and/or its plugins, you might break unrelated pieces of software. 

Therefore I agree with @reneeotten: only upgrade when required.

upt_macports/upt_macports.py
```diff
@@ -176,6 +180,47 @@ def create_package(self, upt_pkg, output=None):
...
         packager = pkg_cls()
         packager.create_package(upt_pkg, output)
+
+    def pkg_exist(self):
+        prefix = {
+            'pypi': 'py-',
+            'rubygems': 'rb-',
+            'cpan': 'p5-',
+        }
+        port_name = prefix.get(self.upt_pkg.frontend, '') + self.upt_pkg.name
```
Steap: Don't we already have a _normalized_macports_name static method that determines the port name? Or does it do something slightly different?

Anyway, it would be great if this were handled by a static method in the subclasses of MacPortsPackage.

upt_macports/upt_macports.py
```diff
@@ -176,6 +180,47 @@ def create_package(self, upt_pkg, output=None):
...
+        self.logger.info(f'Checking MacPorts tree for port {port_name}')
+        cmd = f'port info --version {port_name}'
+        port = subprocess.getoutput(cmd)
```
Steap, line 260: Does port(1) never fail? Isn't there some error-handling we'd like to do here?

upt_macports/upt_macports.py
```diff
@@ -176,6 +180,47 @@ def create_package(self, upt_pkg, output=None):
...
+        self.logger.info(f'Current Upstream version is {self.upt_pkg.version}')
+        if port[:5] == 'Error':
+            self.logger.info(f'{port_name} not found in MacPorts tree')
+            self.curr_ver = ''
```
Steap: We are never going to re-use this, right?

And even in the "else:" part of the code, it feels like you are adding info to attributes only so that it can be used in the caller. This makes the code hard to read, as it becomes "spaghetti code", I think.

Maybe pkg_exist could return a tuple (such as (boolean:exist, str:version))? Or just the version: a string when the package exists, None otherwise? Of course the name of the method should be updated.

upt_macports/upt_macports.py
```diff
@@ -176,6 +180,47 @@ def create_package(self, upt_pkg, output=None):
...
+            return False
+        else:
+            self.curr_ver = port.split()[1]
+            self.logger.info(
+                f'Current MacPorts Version for {port_name} is {self.curr_ver}')
+            return True
+
+    def check_needed(self, upt_pkg, constraint):
```
Steap: This whole method is difficult to read, in my opinion.

Don't do:

```python
if X:
    # Some
    # Long
    # Code
    # For
    # The
    # Most
    # Common
    # Case
else:
    return Y
```

Instead, write:

```python
if not X:
    return Y

# Some
# Long
# Code
# ...
```

This way the reader does not have to remember as much context when reading the "real" code.

upt_macports/upt_macports.py
```diff
@@ -176,6 +180,47 @@ def create_package(self, upt_pkg, output=None):
...
+            curr_ver = self.curr_ver
+            upstream_ver = self.upt_pkg.version
+            # TODO: Add Decision Logic
+            if constraint:
+                self.logger.info(f'Constraint is {constraint}')
+            else:
+                self.logger.info('No constraint')
+            if curr_ver < upstream_ver:
+                self.logger.info(
+                    'Creating Portfile as new version is available')
```
Steap: None of these error messages are really useful if you do not include the name of the package.

upt_macports/upt_macports.py
Korusuke: Even if `port` fails then that output will be stored in port which the case when port is not found:
`Error: Port portname not found`

reneeotten: I assume that @Steap refers to the actual ```port``` command failing, for example because it is not installed, not in the path or something similar.

Korusuke: ohh ok, then for that we can do something like:
```
if port.startswith('Error') 
 ...
elif port.startswith('version'):
 ...
else:
    `throw_an_error`
```

reneeotten: Exactly, we should issue a Warning/Error that we cannot run the ```port``` command

upt_macports/upt_macports.py
```diff
@@ -173,9 +178,38 @@ def create_package(self, upt_pkg, output=None):
...
+        try:
+            port_name = pkg_classes[self.frontend](
+            )._normalized_macports_name(name)
+        except KeyError:
+            raise upt.UnhandledFrontendError(self.name, self.upt_pkg.frontend)
+
+        self.logger.info(f'Checking MacPorts tree for port {port_name}')
+        cmd = f'port info --version {port_name}'
+        port = subprocess.getoutput(cmd)
+        # TODO: version not found when macports tree is not updated
```
reneeotten: I don't understand this TODO message. If the MacPorts tree is not up-to-date there are two options: (i) the port doesn't exist yet in the local tree but is available in the GitHub tree or (ii) the version that will be returned is lower if the port has been updated already in the GitHub tree. So I don't think there is anything else left TODO here, or... what are you thinking about here?

upt_macports/upt_macports.py
```diff
@@ -173,9 +178,38 @@ def create_package(self, upt_pkg, output=None):
...
+            self.logger.info(f'{port_name} not found in MacPorts tree')
+            return ''
+        elif port.startswith('version'):
+            curr_ver = port.split()[1]
+            self.logger.info(
+                f'Current MacPorts Version for {port_name} is {curr_ver}')
+            return curr_ver
+        else:
+            sys.exit('Macports tree version detection failed, try running '
+                     'port selfupdate')
```
reneeotten: If I understand it correctly we only end up in this ```else``` block if the ```port``` command failed. So the actual output of the subprocess command is along the lines of ```'/bin/sh: port: command not found'```. I am fine with catching this as with the ```else``` statement, but then the message that will be printed is weird. Shoudn't is just be something like 
```
sys.exit(f'The command "{cmd}" failed. Pleae make sure you have MacPorts installed and/or your PATH is set-up correctly.')
```

reneeotten (commented):
two small questions/suggestions and please rebase with master to resolve the conflicts.

upt_macports/upt_macports.py
Korusuke: So earlier when I was testing recursive packaging, it stopped working without any changes in code which was dues to change in output of `port info` as MacPorts tree was not updated so the output was:
```
Warning: port definitions are more than two weeks old, consider updating them by running 'port selfupdate'.
version: 0.1.2
```
because of this I had added the TODO which still is not handled properly.
We can either use the local version or sys.exit if such Warning is there.

Korusuke: Now this is solved by changing second condition to
```
 elif 'version' in port:
            curr_ver = port.split('version: ')[1]
```

upt_macports/upt_macports.py
Korusuke: Yeah I was supposed to change that message after dealing with the TODO, its updated now.

upt_macports/upt_macports.py
reneeotten: ah, good point - I guess I never go two weeks without updating the port definitions ;) Then just add another ```elif``` statement to catch this and emit a proper message right now instead of adding a TODO statement.

reneeotten: > Now this is solved by changing second condition to
> 
> ```
>  elif 'version' in port:
>             curr_ver = port.split('version: ')[1]
> ```

I don't think that is a good idea as the MacPorts warning of having a >2 weeks outdated ports tree will get buried in this case. As I said, I would handle this as a separate condition and when we run into this situation warn the user about his/her outdated ports tree and tell them to resolve that first (i.e., to a ```sys.exit``` here). Many ports will have been updated within this two week period and working with an outdated tree is never a good idea when trying to package stuff. In addition, we should add in the (still to be written) documentation and/or README that one is assumed to have an up-to-date, local MacPorts tree.

ghosthound: In general use I agree with not proceeding if the ports are outdated, however there are times when that's appropriate (like trying to keep a system at a static set of ports but you need to add another).  If we can include an override (maybe -f is good enough?) so that it is possible to run this on an "outdated" tree I think that'd solve the use case.

reneeotten: thanks for the feedback @ghosthound. 

Adding an override flag would likely need to be done in ```upt``` so that would require some additional considerations (e.g., is this a general enough use-case for other backends or something specific to MacPorts, are there more things that one could/should override with the ```-f``` flag, etcetera).

I am not necessarily set on doing a ```sys.exit``` and *requiring* the user to update the ports tree (although this would be in general the preferred action). We could consider to clearly log that "the tree is outdated and we recommend to update the tree first", and then leave it to the user to decide. But for sure, I would rather not ignore the warning message of an outdated tree as currently happens.

Korusuke:
I guess we can merge this now as @Steap has already added Recursive feature in upt core.

reneeotten:
> I guess we can merge this now as @Steap has already added Recursive feature in upt core.

I will install the latest ```upt``` and the code from this PR and give it a last test today/tomorrow; if everything works as intended it can indeed be merged.

Korusuke:
While writing the update feature, there are a few things that I guess will be changed in package_versions() so please don't merge this as I would like to change those here itself...will commit those changes tomorrow once I am sure of it.

upt_macports/upt_macports.py
```diff
@@ -235,9 +239,46 @@ def create_package(self, upt_pkg, output=None):
...
         packager = pkg_cls()
         packager.create_package(upt_pkg, output)
+
+    def package_versions(self, name):
+        pkg_classes = {
+            'pypi': MacPortsPythonPackage,
+            'cpan': MacPortsPerlPackage,
+            'rubygems': MacPortsRubyPackage,
+            'npm': MacPortsNpmPackage
+        }
```
Steap: Please don't do this. This "dispatch" thing should only be done once.

reneeotten:
okay, I did some testing and to from a user-perspective it all seems to work as intended - very cool!

I am not the biggest fan of the messages for checking whether to package something or not; in MacPorts there is always only one version available. But, I understand that this is needed in ```upt``` for other backends and I don't think it's worthwhile to override the ```def needs_requirement(self, req, phase):``` just for some more accurate or (for me personally) more pleasing messages ;)

```[INFO    ] [Backend] Dependency numpy (>=1.0): at least one compatible version found in ['1.16.4']. Not packaging it.```

There are two "unresolved" (related) conversation that were brought up by @Steap, so let's try to fix that. We'll wait for what other changes you think are needed before merging. Again, please do NOT force-push to this branch, just add separate commits for now; it makes it easier to review the changes and if needed we can squash things before merging the PR.

**Comment by Steap** [comment]
I hadn't thought of that :) It is true that it feels unnatural to have a list for MacPorts, but I guess it's a minor inconvenience :)

I agree with the idea of squashing. It's too bad github cannot remember multiple commits like Gerrit does, for instance.

File: upt_macports/upt_macports.py [hunk 16]
```diff
@@ -229,32 +229,25 @@ def __init__(self):
...
-        pkg_classes = {
-            'pypi': MacPortsPythonPackage,
-            'cpan': MacPortsPerlPackage,
-            'rubygems': MacPortsRubyPackage,
-            'npm': MacPortsNpmPackage
-        }
         try:
-            port_name = pkg_classes[self.frontend](
+            port_name = self.pkg_classes[self.frontend](
             )._normalized_macports_name(name)
```
Steap: You are instantiating a new upt.Package  object. Why remember self.pkg_classes? Why not keep a handle on "packager" in create_package()?

Korusuke: Sorry, but I didn't get you.

reneeotten: I am not sure either; @Steap can you educate us? ;)

Anyway, let's resolve this specific issue (#41) right now and revisit the other point (#43,  about where/how to centralize the naming of ports/directories/dependencies) in a later PR. We should really merge this PR soon so that we can move on to the "update" PR - almost there!

Korusuke: Yeah agree, also we should address the [version issue](https://github.com/macports/upt-macports/issues/44#issuecomment-513349606) here.

I'll push the changes in few min.

reneeotten: I agree that this should be fixed (haven't looked carefully at it though), but if you're description is correct it's not something we should fix here. That sounds like an issue with MacPorts not with the code you have, or...? 

If I am correct about this, then it has nothing to do with this PR, so let's not push changes to this PR.

Korusuke: yeah ok.

Korusuke: This changes should work perfectly now, the changes I mentioned before that I did when adding update feature is just having `package_versions` take an extra argument as `frontend name` but since upt-core has already merged this part of code so we should not change it now....we can change it while merging update feature instead.

Steap: When you write:

```python
 port_name = self.pkg_classes[self.frontend]()._normalized_macports_name(name)
```
you are creating a upt.Package() instance that you use only once and then throw away.

You could probably keep the package object by using it as an attribute (self.package = self.pkg_classes[self.frontend]() in create_package()).

Steap:
Commit "Not scanning for frontend when creating output dir " should probably be a separate PR. It would be fairly simple to review and merge, and would in turn make this PR easier to review.

upt_macports/upt_macports.py
Korusuke: If I understand this correctly then this will only work if `create_package` is called before `package_versions` or `update_package`(in case of update).

But as you can see in upt-core for update feature we will call `package_versions` before `update_package` and we actually never call `create_package`

Korusuke:
> Commit "Not scanning for frontend when creating output dir " should probably be a separate PR. It would be fairly simple to review and merge, and would in turn make this PR easier to review.

Moved to another PR

upt_macports/upt_macports.py
Steap: By the way, I meant "MacPortsPackage", not "upt.Package", sorry :)

Hmm, okay, that's a good point. Maybe we'll change Backend.__init__() to add frontend as an argument, which might be the cleanest way of doing this, but I'm not sure. I'll have to think about this :)

Korusuke: So should we get this PR merged now as it's in working state and we can change this later on?

reneeotten: Yes, that would be fine with me! @Steap can we work on finalizing PR 47 and this one today? That would make reviewing the "update PR" easier since that has several of these commits mixed in..

upt_macports/upt_macports.py
```diff
@@ -258,7 +258,7 @@ def create_package(self, upt_pkg, output=None):
     def package_versions(self, name):
         try:
             port_name = self.pkg_classes[self.frontend](
-            )._normalized_macports_name(name)
+            )._normalized_macports_folder(name)
```
Steap: So, can we avoid creating a new object here? Just like in PR #47 .

upt_macports/upt_macports.py
Steap: @Korusuke Can you rebase this on top of master, and test it with the latest commits of upt? I released 0.10.1 yesterday.

Korusuke: Yep, it works perfectly!

upt_macports/upt_macports.py
Korusuke: Is it fine now?

upt_macports/upt_macports.py
```diff
@@ -227,20 +228,53 @@ def jinja2_reqformat(self, req):
...
+    def __init__(self):
+        self.logger = logging.getLogger('upt')
+
     name = 'macports'
+    pkg_classes = {
+                'pypi': MacPortsPythonPackage,
+                'cpan': MacPortsPerlPackage,
+                'rubygems': MacPortsRubyPackage,
+                'npm': MacPortsNpmPackage
+            }
```
Steap: That is a weird indentation style. I think it's more common to do:

```python
pkg_classes = {
    'pypi': MacPortsPythonPackage,
    ...
}
```

Steap:
So this patch looks quite good now.

@reneeotten What do you think, especially about the MacPorts-specific parts?

Should @Korusuke squash the commits, rebase them (if needed) and show us a "final" patch?

reneeotten:
> So this patch looks quite good now.
> 
> @reneeotten What do you think, especially about the MacPorts-specific parts?
> 
> Should @Korusuke squash the commits, rebase them (if needed) and show us a "final" patch?

It looks good to me; just tested a little bit again and it works as intended. @Korusuke please change the indentation and squash the commits as @Steap suggested and then let's merge this!

reneeotten:
@Steap I think we should merge this and will do so in a few hours unless you have objections. If you agree, feel free to merge it yourself at your earliest convenience ;)

PR #37, opened 2019-06-27
Fix ruby Portfile

### LINKED ISSUE #36: Ruby ports
I just added Ruby 2.6 as supported version to the Ruby PortGroup (https://github.com/macports/macports-ports/commit/b4b32250f11c731de991bd6b51a4c626b3eed9ad). 

After doing so I tried to package a (random) port (```rb-treetop```) to see how ```upt``` performs. It does get all the dependencies and such right, but there are a few issues with the current template. Admittedly, we (or probably I) will need to do some more work on the PortGroup itself to fix some of them.... but that might take a while since I don't have much experience with Ruby nor writing PortGroups ;)

Anyway, the template will eventually need some more changes once the new PortGroup is ready, but for now it should result in a working Portfile with the current PG which isn't the case right now. 

Specifically:
1. ```ruby.branches``` needs to be before ```ruby.setup```, otherwise there are no subports generated (I would consider this a bug in the PG)
2. for dependencies it shouldn't be ```port:rb${ruby.version}-<name>``` but ```port:rb${ruby.suffix}-<name>```

Trying to package its dependency (```rb-polyglot```) there is an issue with the formatting of the ```description``` that makes the Portfile fail the parsing by MacPorts. Fixing that makes both ports (as far as I can tell at least) install correctly. 

Finally, it logs the following message (twice):
```[ERROR   ] [Backend] Could not determine the type of the source archive```
That should not happen I think as the source should be always ```.gem``` here, but also with the latest release of ```upt-rubygems``` I thought is was setting an ArchiveType, correct? You made an exception for rubygems in the ```base.Portfile``` template for the archive_type, perhaps the same thing should happen for the logging as well?

### PR DESCRIPTION
**Pull request by Korusuke**
Fixes #36

### DIFF
--- a/upt_macports/upt_macports.py
+++ b/upt_macports/upt_macports.py
@@ -79,7 +79,8 @@ class MacPortsPackage(object):
             'xz': 'xz'
         }
         try:
-            archive_name = self.upt_pkg.get_archive().filename
+            archive_name = self.upt_pkg.get_archive(
+                self.archive_format).filename
             archive_type = archive_name.split('.')[-1]
             return archive_keyword.get(archive_type, 'unknown')
 
@@ -90,6 +91,7 @@ class MacPortsPackage(object):
 
 class MacPortsPythonPackage(MacPortsPackage):
     template = 'python.Portfile'
+    archive_format = upt.ArchiveType.SOURCE_TARGZ
 
     def _pkgname(self):
         macports_name = self._normalized_macports_name(self.upt_pkg.name)
@@ -121,6 +123,7 @@ class MacPortsNpmPackage(MacPortsPackage):
 
 class MacPortsPerlPackage(MacPortsPackage):
     template = 'perl.Portfile'
+    archive_format = upt.ArchiveType.SOURCE_TARGZ
 
     def _pkgname(self):
         macports_name = self._normalized_macports_name(self.upt_pkg.name)
@@ -157,6 +160,7 @@ class MacPortsPerlPackage(MacPortsPackage):
 
 class MacPortsRubyPackage(MacPortsPackage):
     template = 'ruby.Portfile'
+    archive_format = upt.ArchiveType.RUBYGEM
 
     def _pkgname(self):
         macports_name = self._normalized_macports_name(self.upt_pkg.name)
--- a/upt_macports/templates/base.Portfile
+++ b/upt_macports/templates/base.Portfile
@@ -24,7 +24,7 @@ license             {{ pkg.licenses }}
 # maintainers         {example.org:julesverne @github-handle} openmaintainer
 maintainers         nomaintainer
 
-description         {{ pkg.upt_pkg.summary }}
+description         {{ pkg.upt_pkg.summary|trim|replace('\n',' \\\n')|indent(20, false) }}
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by reneeotten** [comment]
thanks @Korusuke the changes in the template look good and work as expected! 

I just had another look at the error message ``` [ERROR ] [Backend] Could not determine the type of the source archive``` that you now successfully suppress, but on second thought there shouldn't be a need for that.

@Steap added Archive to ```upt-rubygems``` and the frontend indeed correctly returns an archive. I think that there is a bug in the ```get_archive``` function in ```upt``` itself. If you add some print statements and run ```upt package -f rubygems -b macports treetop``` then you'll see the following:
```
self.archives = [<upt.upt.Archive object at 0x106664c50>]
archive.archive_type = ArchiveType.RUBYGEM
```
which seems all correct. However, ```archive_type``` in this function is still ```archive_type = ArchiveType.SOURCE_TARGZ``` and that means that it's filtered out in lines 187-188.

**Comment by Korusuke** [comment]
Yes, agree even I think that it should be fixed in upt as `get_archive()` sets the archive type as `targz2` by default.
As of now I have changed the code so it makes a exception while calling the function and not in error as that is more appropriate this way but I still think that this should be handled by `upt`.

**Comment by reneeotten** [comment]
I guess my choice words using "bug" is probably too strong / wrong, it might actually be a feature ;) In all cases except for ```upt-rubygems``` the ```SOURCE_TARGZ``` is the correct type, so it makes sense that this is the default. So perhaps your change to only call the ```_get_archive``` function with specifying the ```RUBYGEM``` archive type is the correct thing to do, I am not sure whether ```upt``` can take care of this automagically - lets see what @Steap has to say.

File: upt_macports/upt_macports.py [hunk 1]
```diff
@@ -79,7 +79,12 @@ def archive_type(self):
             'xz': 'xz'
         }
         try:
-            archive_name = self.upt_pkg.get_archive().filename
+            if self.upt_pkg.frontend != 'rubygems':
+                archive_name = self.upt_pkg.get_archive().filename
+            else:
+                archive_name = self.upt_pkg.get_archive(
+                    self.upt_pkgArchiveType.RUBYGEM).filename
+
```
Steap: What you could do here is define an "archive_type" class attribute for all subclasses of MacPortsPackage (just like the "template" attribute) that would be equal to "upt.ArchiveType.RUBYGEM" or "upt.ArchiveType.SOURCE_TARGZ" (note that the latter does not necessarily refer to a tar.gz, because I fucked up when naming it :p).

You would then write:

```python
archive_name = self.upt_pkg.get_archive(self.archive_type).filename
```

Steap:
> lets see what @Steap has to say.

Yep, you're right. get_archive() defaults to looking for a source archive, because that is the most common use case, but one may want to look for a Python wheel, a Ruby gem, etc.

reneeotten:
thanks @Korusuke I am happy with this, it works exactly as intended now! @Steap if you agree, please do merge this PR or approve the changes and I'll get to it tomorrow.

Steap:
@reneeotten It seems OK, but I'll let you merge, you understand these Ruby/Macports issues better than me :)

PR #40, opened 2019-07-02
More tests and codecov Integration

Korusuke:


reneeotten:
this looks good to me and ready to merge - @Steap?

Steap:
@reneeotten I don't really know codecov, so I'll let you merge this :)

Does it deprecate the use of coverage (see tox.ini)?

reneeotten:
You still need to run coverage, ```codecov``` will tell you how your coverage changes with a PR and you can easily check which lines are not executed with the tests. It can be quite useful and certainly doesn't hurt!

I suppose the MacPorts repository is already set-up correctly since also ```macports-webapp``` is using it. I am going to merge this and then we'll find out...

PR #47, opened 2019-07-20
Not scanning for frontend when creating output dir

### LINKED ISSUE #41: Do not scan the frontend twice
In _create_output_directories, we "scan" the frontend:

```python
    upt2macports = {
        'cpan': 'perl',
        'pypi': 'python',
        'rubygems': 'ruby',
    }
``` 

We already do that once in MacportsBackend.create_package(). We should only do it once.

Korusuke:
Fixes #41


Korusuke:
I don't think there might be anything wrong in this PR as this was previously disscussed.
So we can merge this first and then rebase and merge #35

reneeotten:
this looks fine to me. Since @Steap opened the issue about this I'll let him take another look and, if he agrees, merge this PR!

Steap:
Done :)

PR #51, opened 2019-08-08
cpan version conversion

Korusuke:


upt_macports/upt_macports.py
```diff
@@ -278,3 +278,32 @@ def package_versions(self, name):
             sys.exit(f'The command "{cmd}" failed. '
                      'Please make sure you have MacPorts installed '
                      'and/or your PATH is set-up correctly.')
+
+    def cpan_version(self, vers):
```
reneeotten: the function itself works as intentded, at least to the same extent as what's happening in MacPorts (more on that below). However, the code looks a bit "ugly" - not very Pythonic and it's probably a direct "translation" of the Perl procedure. Perhaps we can improve on that a little.

upt_macports/upt_macports.py
```diff
@@ -278,3 +278,32 @@ def package_versions(self, name):
...
+            sub = fractional[index:index+3]
+            if len(sub) < 3:
+                sub += '0'*(3-len(sub))
+            ret = ret + '.' + str(int(sub))
+            index += 3
+        return ret
+
+    def needs_requirement(self, req, phase, frontend):
+        if frontend == 'cpan' and req.specifier:
+            print(req.specifier)
```
reneeotten: in the "final" version we don't really need a print-statement anymore - correct?

upt_macports/upt_macports.py
```diff
@@ -278,3 +278,32 @@ def package_versions(self, name):
...
+            req.specifier = '>=' + self.cpan_version(req.specifier[2:])
```
reneeotten: I don't think this can be generalized like this. Why do you assume the specifier always start with ```>=```? The way I read [the CPAN information on version ranges](https://metacpan.org/pod/CPAN::Meta::Spec#Version-Ranges), there are several options, including only a version number, or for example a one character operator like ```>```. 

I am not even sure if ```upt``` parses more complicated specifiers like ```>= 1.2, != 1.5, < 2.0``` but if so we would need to take that into account as well (perhaps also in the ```needs_requirement``` function, I don't know). So perhaps we need to use a regular expression here to separate the version from the operator.

reneeotten (commented):
a few comment inline

upt_macports/upt_macports.py
Steap: According to your link, CPAN has <, <=, ==, >=, > and !=. These are all supported by upt, which uses [PEP440](https://www.python.org/dev/peps/pep-0440/) internally.

If CPAN specifies "2.0", it means ">=2.0", and it's not properly parsed by upt-cpan, I'll need to fix this :)

upt_macports/upt_macports.py
Steap: Yeah, I'd love to know whether you "converted" this from Perl code, and if so, if you could give us a link to this code.

This is also the perfect patch for TDD: write all your test cases, then make sure your function works.

reneeotten:
The ```cpan_version``` function does the exact same conversion as MacPorts currently does, so comparing that to ```port info --version <port>``` for the "recursive packaging" and "update" using ```upt``` should work. 

However, compared to the recommended-Perl-way of doing the conversion (using ```perl -Mversion -e 'print version->parse(*version*)->normal'```) there are some discrepancies that are probably worthwhile to investigate and perhaps correct both here as in the Perl PG procedure. 

For example, versions that do not contain a ```.``` are not converted correctly compared to the Perl-way:
```
p5-perl-tidy    20190601 (portfile) ; 20190601 (port info --version) ; v20190601.0.0 (Perl) ; 20190601 (cpan_version function)
```
other ports in this category are: p5-test-regexp, p5-canary-stability, p5-x11-protocol-other, p5-uri-find, p5-test-perltidy, p5-regexp-common, p5-time-y2038, p5-mozilla-ca

@mojca @dbevans (or anybody else following this): does any of you remember if there was/is a reason for this?

Then there are some others that I have not checked yet carefully that do not give the same result compare to the Perl conversion, but do agree with ```port info --version```:
p5-nkf, p5-ifeffit, p5-business-isbn-data, p5-postscript-font, p5-mime-charset, p5-mime-encwords, p5-gearman

Some of these might just be mistakes in the Portfile or easy to fix unrelated problems so let's not worry about them you.... I'll look at some of those ports.

upt_macports/upt_macports.py
reneeotten: I understand that the operators are supported by ```upt```. My example specifier ```>= 1.2, != 1.5, < 2.0``` means that all version equal/above 1.2 and below 2 with the exception of 1.5 are supported. The question is whether the function ```needs_requirements``` in ```upt``` considers all these three constraints in the specifier when doing:

```
s = SpecifierSet(req.specifier)
compatible_versions = list(s.filter(versions))
```
, where versions is the list of versions that are packaged by the backend. I haven't checked that, but I was hoping one of you could save me the time trying ;)

upt_macports/upt_macports.py
Korusuke: It's a direct translation of `proc perl5_convert_version {vers} ` from [perl portgroup](https://github.com/macports/macports-ports/blob/master/_resources/port1.0/group/perl5-1.0.tcl). 

Yeah, right...I'll add the tests.

upt_macports/upt_macports.py
Korusuke: Ohh sorry.

upt_macports/upt_macports.py
reneeotten: Sorry I meant to say the Tcl procedure... the code for converting "upstream" versions to what ```port info --version <package>``` returns (and is supposed to follow the Perl conversion) is part of the perl5 PortGroup. The file is found in the ```macports-ports``` repository under ```_resources/port1.0/group/perl5-1.0.tcl```. 

The relevant code is:
```
# Convert a floating-point version to an equivalent dotted decimal one.
# If version is expressed as a perl v-string, strip the leading "v"
proc perl5_convert_version {vers} {
    if {[string index $vers 0] eq "v"} {
        set start 1
    } else {
        set start 0
    }
    set index [string first . $vers]
    set other_dot [string first . [string range $vers [expr {$index + 1}] end]]
    if {$index == -1 || $other_dot != -1} {
        return [string range $vers $start end]
    }
    set ret [string range $vers $start [expr {$index - 1}]]
    incr index
    set fractional [string range $vers $index end]
    set index 0
    while {$index < [string length $fractional] || $index < 6} {
        set sub [string range $fractional $index [expr {$index + 2}]]
        if {[string length $sub] < 3} {
            append sub [string repeat "0" [expr {3 - [string length $sub]}]]
        }
        append ret ".[scan $sub %u]"
        incr index 3
    }
    return $ret
```

Korusuke:
> However, compared to the recommended-Perl-way of doing the conversion (using `perl -Mversion -e 'print version->parse(*version*)->normal'`) there are some discrepancies that are probably worthwhile to investigate and perhaps correct both here as in the Perl PG procedure.
> 
> For example, versions that do not contain a `.` are not converted correctly compared to the Perl-way:
> 
> ```
> p5-perl-tidy    20190601 (portfile) ; 20190601 (port info --version) ; v20190601.0.0 (Perl) ; 20190601 (cpan_version function)
> ```

I looked into this a bit and from what I understand, perl conversion gives 'v20190601.0.0' but since there are no decimal points we don't need to worry about any issue of version comparison mismatch for which we are actually doing all this. So as per me we don't need to match everything to "perl conversion method", also because in macports v-string as version is not allowed as far as I know.

Also I guess since we are doing this for comparing versions we should cross check the conversion with:
`perl -Mversion -e 'print version->parse(*version*)->numify`
instead of
`perl -Mversion -e 'print version->parse(*version*)->normal`

Doing this doesn't change @reneeotten analysis but the output changes a bit for most ports like for p5-perl-tidy now we have `20190601.000` instead of `v20190601.0.0`.

The perl conversion for `p5-business-isbn-data` is very weird:
```
using normal -> v20140910.2.999.999
using numify -> 20140910.002999999
```

Korusuke:
> Also I guess since we are doing this for comparing versions we should cross check the conversion with:
> `perl -Mversion -e 'print version->parse(*version*)->numify`
> instead of
> `perl -Mversion -e 'print version->parse(*version*)->normal`

I tried this out for all ports and I guess we should follow this only if we change it in portgroup as 'numify' changes things quite a lot for the ports where `port --info version` and `perl normal` were same.

reneeotten:
@Korusuke I am not suggesting anywhere to change the conversion from "normal" (that is currently done) to "numify" in MacPorts. Chances for that to happen are probably not high anyway and I also don't see an obvious reason to do so. And, yes, of course ```port info``` will not give versions that start with a ```v```, but I'm not sure that is observation is relevant here.

The only thing I pointed out is that the current MacPorts conversion isn't fully consistent with (the presumably intended) conversion of ``` perl -Mversion -e 'print version->parse(*version*)->normal```. That's -as you also pointed out- not an issue as long as we do the exact MacPorts conversion in also in the  ```upt_macports``` backend. Eventually, if this conversion ends up in ```upt``` this is something to reconsider; of course, we can always just keep it the same and overwrite the ```needs_requirement``` functions as you've done in this PR. 

So, there is no real reason to spend time on investigating this right now, let's instead fix/improve the conversion function itself and get the code in ```needs_requirement``` working correctly... or?

jmroot:
> The perl conversion for `p5-business-isbn-data` is very weird:
> 
> ```
> using normal -> v20140910.2.999.999
> using numify -> 20140910.002999999
> ```

That would be a floating-point precision issue. (Which is one of the very good reasons not to use floating-point numbers for things that need to be compared exactly, especially when converted back and forth between binary and decimal representations.)

upt_macports/tests/test_macports_backend.py
```diff
@@ -49,5 +49,19 @@ def test_port_error(self, mock_sub):
...
 
+class TestMacPortsCpanVersion(unittest.TestCase):
+    def setUp(self):
+        self.macports_backend = MacPortsBackend()
+        self.macports_backend.upt_pkg = upt.Package('foo', '42')
+        self.macports_backend.frontend = 'cpan'
+
+    def test_version_conversion(self):
+        converted = ['1', '1.2.3', '1.200.0', '1.200.0', '1.20.0']
+        upstream = ['1', '1.2.3', '1.2', '1.20', '1.02']
```
reneeotten: thank you for adding tests! In the conversion code (both in MacPorts and here), a possible leading ```v``` in the upstream version is stripped in the function before doing the conversion. Nevertheless, you will be sometimes passing in an upstream version with a ```v``` present and you probably want to check that the conversion works correctly in such cases as well

upt_macports/upt_macports.py
reneeotten: > If CPAN specifies "2.0", it means ">=2.0", and it's not properly parsed by upt-cpan, I'll need to fix this :)

Right, if CPAN specifies "2.0" that indeed means "=>2.0" but the conversion in the current implementation will do
```req.specifier = '>=' + self.cpan_version(req.specifier[2:])``` and result in ```req.specifier``` returned being ```>=0```.

Additionally, if the specifier contains multiple entries as in my example below; then we should do the version conversion on all of these specifiers.

Comment (reneeotten) [reply upt_macports/upt_macports.py]: >> According to your link, CPAN has <, <=, ==, >=, > and !=. These are all supported by upt, which uses [PEP440](https://www.python.org/dev/peps/pep-0440/) internally.

okay, I just tried and indeed the more complicated specifiers do work correctly. I didn't look carefully enough at the ```filter``` function, so that's all good. The comment above still holds true though.

Comment (Steap) [reply upt_macports/upt_macports.py]: Oh, I see your point, you make perfect sense. By the way, I think a similar issue exists in upt-cpan.

Would you know of a Perl package that has a "complex" specifier, such as the one from the documentation?

Comment (reneeotten) [reply upt_macports/upt_macports.py]: @Steap a few examples are:
1. specifier with ```<```: Lexical-SealRequireHints under runtime/conflicts
2. with ```>=``` and ```!=```: Net-Twitter and Net-Twitter-Lite under runtime/requires
3. with ```==```: HTML-HeadParser-Liberal under runtime/requires

Comment (Korusuke) [reply upt_macports/upt_macports.py]: Now this should work with all specifiers once it is fixed in upt-cpan

Comment (reneeotten) [reply upt_macports/upt_macports.py]: @Korusuke that change looks good; you should have added tests for this though at the same time ;)

File: upt_macports/upt_macports.py (hunk 1 above)
Comment (reneeotten) [reply upt_macports/upt_macports.py]: why does the function has ```self``` as an argument - it doesn't seem to use it... or? Also did you try to make it more Pythonic instead of a translated Tcl function?

**Comment by Korusuke** [comment]
I guess we are ready to merge this? or are we waiting on upt-cpan to be updated?
About tests for:
```
s = SpecifierSet(req.specifier)
            req.specifier = ','.join(
                [dep.operator + self.cpan_version(dep.version) for dep in s])
```
I tried searching for a way to test the value of variable req but to no luck.
So if we want to really test it I guess we have to put that code in another function which return req, that way it can be tested.
Is there any way I am missing that we can test the variable.

**Comment by reneeotten** [comment]
> I guess we are ready to merge this? or are we waiting on upt-cpan to be updated?

I don't think so yet; there is still the comment on the ```cpan_version``` regarding why it has ```self``` as an argument (but doesn't seem to use it) and I asked you whether you tried to write that function in a more Pythonic way.

Also, what specific issue in ```upt-cpan``` would we be waiting for to be resolved (ping @Steap)? The issue you opened in that repository (https://framagit.org/upt/upt-cpan/issues/2) seems not relevant here. Or perhaps you're thinking about the code below from ```upt_cpan.py```:
```
requirements[upt_phase] = [
    upt.PackageRequirement(module,
                           '' if spec == '0' else f'>={spec}')
    for (module, spec) in dependencies.get('requires', {}).items()
    if module != 'perl'  # This seems implicit
]
```
where it either seems that there is no specific version requirement (```'' if spec == '0'```) or always assumes that the operator is ```>=```? The latter one might indeed not be correct as we discussed earlier above. If that's what you're referring to that would be an issue/PR that belongs in that repository.

File: upt_macports/upt_macports.py [hunk 5]
```diff
@@ -278,3 +279,28 @@ def package_versions(self, name):
...
+        index = 0
+        while index < len(fractional) or index < 6:
+            sub = fractional[index:index+3]
+            if len(sub) < 3:
+                sub += '0'*(3-len(sub))
+            ret = ret + '.' + str(int(sub))
+            index += 3
+        return ret
+
+    def needs_requirement(self, req, phase, frontend):
```
reneeotten: shouldn't the function that you want to override have (ideally) the same number of arguments? I was (unsuccessfully) trying to get the testing to work and run into trouble with "3 versus 4 arguments" - I am not sure if that's completely related though.... In this case, is ```frontend``` really needed, can it not just be ```if self.frontend == cpan``` (it seems to me that this is set already in the ```create_package``` function before?

What I was trying to do is to make a test where you define a requirement with a more complicated specifier and then call the ```needs_requirement``` function of ```MacPortsBackend```. You've been using ```mock``` than me, but wouldn't you be able to see then how ```upt.Backend.needs_requirement``` would get called?

Korusuke: > shouldn't the function that you want to override have (ideally) the same number of arguments? I was (unsuccessfully) trying to get the testing to work and run into trouble with "3 versus 4 arguments" - I am not sure if that's completely related though.... In this case, is `frontend` really needed, can it not just be `if self.frontend == cpan` (it seems to me that this is set already in the `create_package` function before?

This should be fixed now, I was trying this out with update feature since it is easier to test with update feature.
In update feature the extra 'frontend' is required since there self.frontend is not set.

> What I was trying to do is to make a test where you define a requirement with a more complicated specifier and then call the `needs_requirement` function of `MacPortsBackend`. You've been using `mock` than me, but wouldn't you be able to see then how `upt.Backend.needs_requirement` would get called?

That's an awesome idea, tests added for the same.

Korusuke:
> The latter one might indeed not be correct as we discussed earlier above. If that's what you're referring to that would be an issue/PR that belongs in that repository.

Yes I was referring to that, I am under the impression that @Steap is working on both the upt-cpan issues.

> I asked you whether you tried to write that function in a more Pythonic way.

I changed it out a bit from the first version I had written but from the current state, I don't see any particular changes.

upt_macports/tests/test_macports_backend.py
```diff
@@ -64,6 +64,25 @@ def test_version_conversion(self):
             self.assertEqual(
                 self.macports_backend.cpan_version(mp_ver), cpan_ver)
 
+    @mock.patch('upt.Backend.needs_requirement')
+    def test_require(self, mock_need_req):
```
reneeotten: can we give this a bit clearer name, for example somthing like ```test_needs_requirements```?

upt_macports/tests/test_macports_backend.py
```diff
@@ -64,6 +64,25 @@ def test_version_conversion(self):
...
+            '!=42': '!=42',
+            '==42': '==42',
+            '>= 1.2, != 1.5, < 2.0': '>=1.200.0, !=1.500.0, <2.0.0',
+            '>= 1.2, != 2, < 3.0': '>=1.200.0, !=2, <3.0.0'
+        }
+
+        for key, value in specifiers.items():
+            req = upt.PackageRequirement('bar', key)
+            self.macports_backend.needs_requirement(req, 'fake-phase')
+            self.assertCountEqual(
```
reneeotten, line 82: so here you only check whether the number of elements is the same, correct? Can that be expanded to make sure that it actually gives the *exact* expected output as well, you have already the input (as ```key```) and expected output (as ```value```) listed in ```specifier```?

Korusuke: At first even I thought that it just counts number of Elements as that is what the name 'assertCountEqual' suggests but actually it checks if the elements are equal or not too.

other option is assertListEqual but that needs the order of elements to be same too.

upt_macports/tests/test_macports_backend.py
Korusuke: Done.

upt_macports/tests/test_macports_backend.py
reneeotten: ah, okay - sorry I should have read the documentation of the function ;)

upt_macports/upt_macports.py
```diff
@@ -279,3 +280,29 @@ def package_versions(self, name):
             sys.exit(f'The command "{cmd}" failed. '
                      'Please make sure you have MacPorts installed '
                      'and/or your PATH is set-up correctly.')
+
+    @staticmethod
+    def cpan_version(vers):
```
reneeotten: a slightly different version below, not using ```find``` but just doing ```split``` (and including  docstring). I  still don't like the ```while``` loop and one can separate strings in elements of three using
```
from textwrap import wrap
elements = wrap(string, 3)
```
so possibly that is something that can be used here. Can you give that a try?

```
def standardize_perl_version(version):
    """Parse CPAN version and return a normalized, dotted-decimal form.

    The resulting version is identical to the MacPorts conversion as performed
    by the perl5 PortGroup. It is almost the same as the suggested conversion
    using:
       perl -Mversion -e 'print version->parse("<VERSION>")->normal'
    with the exception of version numbers that do not contain a "dot".

    """
    version_strip = version.lstrip('v')
    version_split = version_strip.split('.')

    # no or two 'dots': no conversion required
    if len(version_split) in (1, 3):
        return version_strip

    # conversion required
    std_version  = version_split[0]
    fractional = version_split[1]

    index = 0
    while index < len(fractional) or index < 6:
        sub = fractional[index:index+3]
        if len(sub) < 3:
            sub += '0'*(3-len(sub))
        std_version = std_version + '.' + str(int(sub))
        index += 3

    return std_version
```

reneeotten:
> > The latter one might indeed not be correct as we discussed earlier above. If that's what you're referring to that would be an issue/PR that belongs in that repository.
> 
> Yes I was referring to that, I am under the impression that @Steap is working on both the upt-cpan issues.

I opened an issue in ```upt-cpan``` for this.

reneeotten:
@Korusuke @Steap I squashed the commits and made a few small modifications to the function that converts the CPAN version. If there are no further comments I intend to merge this on Saturday morning (EST time).

PR #61, opened 2019-12-19
Fix license logging and add messages to the template

reneeotten:
This PR makes the changes below to the license logging and templates:

- "no license found" should be logged as warning, not as info
- add the reasons (i.e., "no upstream license found" and "upt failed to detect the license") to the template as helpful information for the maintainer to figure out why the ```license``` field was set to "uknown" by ```upt```
- use two spaces for inline comments


Steap:
It is nice to have more precise error/warning messages, as it will help us debug issues. I second this change.

PR #62, opened 2019-12-19
Remove master_sites and distname from Python template

reneeotten:
These fields are now set to the default values by the Python PG; see commit: https://github.com/macports/macports-ports/commit/46cbc043fc68208dbb8b7ad20e3fe9dadfd9bd12


Steap:
This looks good to me :)

Steap:
I trust you with this change, you know MacPorts better than I do.

PR #63, opened 2019-12-19
Add homepage to Ruby template (fixes lint error)

reneeotten:
If none specified use the default ```https://www.rubygems.org/gems/${portname}```. This should eventually go into a revamped PG, but for now put it in the template.


reneeotten:
I'll rework this into one "homepage" PR as discussed in #64

PR #64, opened 2019-12-20
Set homepage in templates when needed

reneeotten:
If upstream has set a homepage, and that one is different than what is set by the ```perl5-1.0``` PG, add it to the template. Otherwise let the PG take care of it.

### DIFF
--- a/upt_macports/upt_macports.py
+++ b/upt_macports/upt_macports.py
@@ -120,6 +120,10 @@ class MacPortsPackage(object):
             self.logger.error('Could not determine the type of the source archive') # noqa
             return 'unknown'
 
+    @property
+    def homepage(self):
+        return self.upt_pkg.homepage
+
     def _pkgname(self):
         return self._normalized_macports_name(self.upt_pkg.name)
 
@@ -146,6 +150,14 @@ class MacPortsPythonPackage(MacPortsPackage):
 
     def jinja2_reqformat(self, req):
         return f'py${{python.version}}-{req.name.lower()}'
+
+    @property
+    def homepage(self):
+        homepage = self.upt_pkg.homepage
+        if homepage.startswith('http'):
+            return homepage
+        else:
+            return f'https://pypi.org/project/{self.upt_pkg.name}'
 
 
 class MacPortsPerlPackage(MacPortsPackage):
@@ -189,6 +201,15 @@ class MacPortsPerlPackage(MacPortsPackage):
             self.logger.info('Using fallback location for dist file')
             return f' ../../authors/id/{fallback_dist}/'
 
+    @property
+    def homepage(self):
+        homepage = self.upt_pkg.homepage
+        portgroup_default = 'metacpan.org/pod'
+        if homepage.startswith('http') and portgroup_default not in homepage:
+            return homepage
+        else:
+            return None
+
 
 class MacPortsRubyPackage(MacPortsPackage):
     template = 'ruby.Portfile'
@@ -206,6 +227,14 @@ class MacPortsRubyPackage(MacPortsPackage):
 
     def jinja2_reqformat(self, req):
         return f'rb${{ruby.suffix}}-{req.name.lower()}'
+
+    @property
+    def homepage(self):
+        homepage = self.upt_pkg.homepage
+        if homepage.startswith('http'):
+            return homepage
+        else:
+            return f'https://rubygems.org/gems/{self.upt_pkg.name}'
 
 
 class MacPortsBackend(upt.Backend):
--- a/upt_macports/templates/perl.Portfile
+++ b/upt_macports/templates/perl.Portfile
@@ -10,6 +10,10 @@ revision            0
 {% endblock %}
 
 {% block dist_info %}
+{%- if pkg.homepage -%}
+homepage            {{ pkg.homepage }}
+
+{% endif %}
 {% if pkg.archive_type not in ['gz', 'unknown'] -%} {{ "use_%-15s yes"|format(pkg.archive_type) }}
 
 {% endif -%}
--- a/upt_macports/templates/python.Portfile
+++ b/upt_macports/templates/python.Portfile
@@ -13,7 +13,7 @@ revision            0
 {% endblock %}
 
 {% block dist_info %}
-homepage            {{ pkg.upt_pkg.homepage }}
+homepage            {{ pkg.homepage }}
 {% if pkg.archive_type not in ['gz', 'unknown'] -%} {{ "use_%-15s yes"|format(pkg.archive_type) }}
 
 {% else %}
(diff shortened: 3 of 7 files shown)


Steap:
I know I wrote some "complex things" in the templates (such as the "depends" macro), but I usually like to keep the templates simple for two  reasons:

- it makes them easier to read
- it is easier to test Python code than Jinja2 templates (by which I mean "I can't be bothered to find a proper way of testing templates" :p)

So I would rather do something like this:

```python
class MacportsPackage(object):
    ...
    @property
    def homepage(self):
        return self.upt_pkg.homepage

class MacPortsPerlPackage(MacPortsPackage):
    ...
    @property
    def homepage(self):
        if foo:
            return 'http://...'
        else:
            return None
```

In the template we would write:

```jinja
{% if pkg.homepage %}
homepage            {{ pkg.homepage }}
{% endif %}
```

We could then fix the Rubygems issue the same way, and have a bunch of unit tests for everything.

reneeotten:
@Steap I see your point and do think it makes sense to do some of this in the backend instead of templates.

In fact, I started to add a function to automtically "convert" upstream homepages to ```https``` if we know that the hosts support them (e.g., github.com, rubygems.org, etcetera). That is indeed in the ```MacportsPackage``` already, so let me try and combine the PRs with ```homepage```-related changes into one.

Steap:
Do you have examples of packages that report "http" instead of "https"? I think this change should make its way to the frontends, as all backends would probably like that.

reneeotten:
okay, I will update this PR to do the homepage conversion stuff in the backend and simplify the templates.

> Do you have examples of packages that report "http" instead of "https"? I think this change should make its way to the frontends, as all backends would probably like that.

I agree, I think it would actually make sense to add this to ```upt``` itself as not to duplicate the code in every frontend. Something along the lines I added to ```upt_macports``` would probably work:
```
    def _homepage(self):
        upstream_homepage = self.upt_pkg.homepage

        if upstream_homepage.startswith('https'):
            return upstream_homepage

        https_hosts = ('github', 'readthedocs', 'pypi', 'bitbucket',
                       'metacpan', 'rubygems')
        homepage_split = upstream_homepage.lstrip('http://').split('.')

        if set(homepage_split).intersection(https_hosts):
            return upstream_homepage.replace('http', 'https')
        else:
            return upstream_homepage
```

Below are a few examples of ports for which upstream is using ```http```, but that could use ```https```.
- Python/upt-pypi: bigfloat, bottle-sqlalchemy, cffi, https-parser, python-mingus, py, pycares, s3fs, tox, tweepy
- Perl/upt-cpan: Astro-FITS-Header, Catalyst-Plugin-Authorization-ACL, Code-TidyAll, Crypt-OpenSSL-RSA, Data-Visitor
- Ruby/upt-rubygems: activeldap, audiofile, bones, enum, narray

Steap:
I went for a slightly different implementation, what do you think about https://framagit.org/upt/upt/commit/5409ff3a78b78eb526a109b15df3b9ae8c6d5028 ?

reneeotten:
> I went for a slightly different implementation, what do you think about https://framagit.org/upt/upt/commit/5409ff3a78b78eb526a109b15df3b9ae8c6d5028 ?

Yes, your way of checking if one can use ```https``` is much more general and easier to maintain than my poor-man's-approach of adding "known, good hosts that support https". Looks good to me!

I am flying out for the holidays in a few hours, but will eventually update this PR with a more generalized solution for setting the homepage.

reneeotten:
@Steap how about this?

upt_macports/upt_macports.py
```diff
@@ -207,6 +228,14 @@ def _normalized_macports_folder(name):
...
         return f'rb${{ruby.suffix}}-{req.name.lower()}'
 
+    @property
+    def homepage(self):
+        homepage = self.upt_pkg.homepage
+        if homepage.startswith('http'):
+            return homepage
+        else:
+            return f'https://rubygems.org/gems/{self.upt_pkg.name}'
+
```
Steap, line 238: So, in the case of Ruby, using this default URL is already done in upt-rubygems (see https://framagit.org/upt/upt-rubygems/blob/master/upt_rubygems/upt_rubygems.py#L170). Obviously the maintainer could set the homepage to "http://dead-link.com", but hopefully that does not happen too often :)

upt_macports/upt_macports.py
```diff
@@ -189,6 +201,15 @@ def _cpandir(self):
...
             return f' ../../authors/id/{fallback_dist}/'
 
+    @property
+    def homepage(self):
+        homepage = self.upt_pkg.homepage
+        portgroup_default = 'metacpan.org/pod'
+        if homepage.startswith('http') and portgroup_default not in homepage:
+            return homepage
+        else:
+            return None
```
Steap, line 211: Why not use metacpan.org/release/{pkg} as the default value?

upt_macports/upt_macports.py
```diff
@@ -147,6 +151,14 @@ def _normalized_macports_folder(name):
     def jinja2_reqformat(self, req):
         return f'py${{python.version}}-{req.name.lower()}'
 
+    @property
+    def homepage(self):
+        homepage = self.upt_pkg.homepage
+        if homepage.startswith('http'):
+            return homepage
+        else:
+            return f'https://pypi.org/project/{self.upt_pkg.name}'
```
Steap, line 160: Honestly, I should put this default value in upt-pypi. What do you think?

Steap (commented):
A few inlined comments about default values.

upt_macports/upt_macports.py
reneeotten: sure, that wouldn't hurt! But one should make sure then that all situations where the metadata doesn't contain a valid URL are set to this value.

upt_macports/upt_macports.py
reneeotten: well, that's not the value that is used as default in the Perl portgroup; so it if we decide this should be the default it will have to be changed there as well.

upt_macports/upt_macports.py
reneeotten: right, but currently if the metadata contains something random as ```unknown``` or whatever, it will still return that as URL... that's not what we want. It should be set to a default if it's not a valid URL.

reneeotten:
@Steap any additional comments here? If not, I would like to merge this. Of course it would be helpful to add a default in ```upt-pypi``` and such; once that has happened we can always update the code in ```upt-macports``` accordingly. Also, I will look into the URL you suggest for CPAN packages; if we go for that it will need to be changed in the ```perl``` PortGroup as well. Again we can defer this to a later time I think.

reneeotten:
okay, I am going to merge this now - we can make adjustments if things change in ```upt``` or the frontends.

PR #65, opened 2019-12-20
Update metadata

reneeotten:
This PR makes the following changes:

-  add Python 3.7 and 3.8 Trove classifiers
- set ```author``` field to "The MacPorts Project" (@mojca are you okay with that?). I am not sure what e-mail address to use, suggestions are welcome.
- set ```url``` to the GitHub page 
- add Python 3.8 to the Azure pipelines for testing
- use same license text (still BSD-3) and copyright statement as for ```macports-base```

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-Copyright YEAR      Mojca Miklavec
+Copyright 2019      The MacPorts Project
 
 Redistribution and use in source and binary forms, with or without modification,
 are permitted provided that the following conditions are met:
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -15,11 +15,14 @@ strategy:
       imageName: 'ubuntu-16.04'
       python.version: '3.7'
     py36_macos:
-      imageName: 'macos-10.13'
+      imageName: 'macos-10.14'
       python.version: '3.6'
     py37_macos:
-      imageName: 'macos-10.13'
+      imageName: 'macos-10.14'
       python.version: '3.7'
+    py38_macos:
+      imageName: 'macos-10.14'
+      python.version: '3.8'
 
 pool:
   vmImage: $(imageName)
--- a/setup.cfg
+++ b/setup.cfg
@@ -4,15 +4,17 @@ version = 0.1
 description = MacPorts backend for upt.
 long_description = file: README.md
 long_description_content_type = text/markdown
-author = Mojca Miklavec
-author_email = mojca@macports.org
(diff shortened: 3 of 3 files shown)


mojca:
Yes, definitely.

I just put my name there because this was part of the initial template proof-of-concept which I did back in February during a short hacking session. It makes no sense to keep my contact there, using MacPorts is more meaningful.

LICENSE
```diff
@@ -1,26 +1,26 @@
...
-specific prior written permission.
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions
+are met:
+1. Redistributions of source code must retain the above copyright
+   notice, this list of conditions and the following disclaimer.
+2. Redistributions in binary form must reproduce the above copyright
+   notice, this list of conditions and the following disclaimer in the
+   documentation and/or other materials provided with the distribution.
+3. Neither the name of Apple Inc., The MacPorts Project nor the
```
mojca: Why mention Apple, to start with?

Let's just use a simple 2-clause BSD (we ended up doing that in https://github.com/macports/macports-webapp/blob/master/LICENSE, see https://github.com/macports/macports-webapp/pull/115). Base started at Apple, but there's no need to mention Apple here.

mojca (approved):
I would use a simpler licence. Other than that it looks fine.

LICENSE
reneeotten: that's a good point (I remove Apple from the copyright notice, but missed this instance...). Since all the other ```upt``` things are BSD-3 I think it would make sense to keep this the same (and it has been this since the beginning), or...?

reneeotten:
> It makes no sense to keep my contact there, using MacPorts is more meaningful.

Do you have any suggestions about what would be a good e-mail address to add here instead of yours? I was thinking about the ```devel``` list, but it's a bit inconvenient that one would need to subscribe first...

LICENSE
```diff
@@ -1,4 +1,4 @@
-Copyright YEAR      Mojca Miklavec
+Copyright 2018      The MacPorts Project
```
Steap: I'd say "2019" here :)

reneeotten: right.... I clearly need a holiday break ;)

mojca:
Why do you need an email address? URL should be sufficient and if not, macports-devel sounds reasonable.

reneeotten:
> Why do you need an email address? URL should be sufficient and if not, macports-devel sounds reasonable.

If you supply an ```author```, the ```author_email``` is required - I've already added "macports-devel"

PR #66, opened 2020-02-23
default to python 3.8

Korusuke:
As suggested by @mojca in [#6436](https://github.com/macports/macports-ports/pull/6436)


mojca (approved):
Looks good to me.

This reminds me that we'll want to do something with `pypi2port` at some point as well. Either patch that one or maybe print a message saying that one should switch to `upt` :)

PR #68, opened 2020-06-13
Update "_python_root_name" function

reneeotten:
The latest version of upt-pypi (v0.5.1) returns the proper capitalization of the PyPI package name. So there is no need anymore to reconstruct this ourselves from the archive filename.


Steap:
This seems to make the code easier to read, so I'm all for it :)

PR #69, opened 2021-09-10
Implement the "update" feature.

Steap:


Steap:
@reneeotten @mojca Finally, I wrote this :) You can find the logs of a batch update [here](https://perso.aquilenet.fr/~steap/macports-update/updatelogs/). The *.html files will show you a nice "git diff". Most of the failures are due to the fact that the packages were never on {PyPI,CPAN,rubygems} to begin with. Other failures are usually related to the "upstream name" -> "macports name" translation.

I also wrote a [wrapper script](https://perso.aquilenet.fr/~steap/macports-update/port-update.sh) that allows one to run "port-update.sh python/py-upt/Portfile" in order to not deal with the upt command line itself.

Steap:
@reneeotten @mojca Any chance of reviewing this in the coming weeks? I know it's a big diff, but the link in my previous comment should at least give you an idea of what the feature can achieve, and the wrapper script allows you to try the feature for yourself easily:)

upt_macports/upt_macports.py
```diff
@@ -328,3 +330,34 @@ def needs_requirement(self, req, phase):
...
+        # TODO: This is basically the same code as the one found in
+        # MacPortsPackage._create_output_directories(). It would be nice not to
+        # repeat ourselves.
+        if output is None:
+            pkgname = pdiff.new.name
+            folder_name = macports_pkg._normalized_macports_folder(pkgname)
+            output_dir = os.path.join(macports_pkg.category, folder_name)
+            portfile_path = f'{output_dir}/Portfile'
+        else:
+            portfile_path = output
```
reneeotten, line 360: I am not sure that I see the use-case for specifying an alternative output here... I'd think that if you want to update a port, you just want to overwrite the Portfile that is currently being used - or? Also, if you specify an output here I would expect it to behave the same as when not using the `-u` flag. In other words if you specify `-o ~/my-macports-clone` it will generate the Portfile in `~/my-macports-clone/python/py-<packagename>/Portfile`. It behaves quite differently here in the sense that you'd need to supply a full path including the `Portfile` part.

reneeotten (commented):
@Steap I just started to take a quick look and when trying it I got confused about the `-o` flag. I am sure you have a reasoning for it, it's just that I am not completely seeing it yet ;) More to follow...

upt_macports/upt_macports.py
Steap: Hey, thanks for taking a look at the patch!

I'll use PyPI-related examples in my comments, but Rubygems/CPAN work similarly.


> I'd think that if you want to update a port, you just want to overwrite the Portfile that is currently being used - or? 

Yes. The current Portfile is both an input (read the "old" values) and an output (it is overwritten with up-to-date info). upt-macports need to find where the Portfile is, and usually it is really simple: for instance, if a package from PyPI named "requests" must be updated, then the current Portfile must be ./python/py-requests/Portfile. But sometimes, because our naming policy is not bulletproof, this automated "PyPI name to Macports name" does not work, leaving upt-macports unable to find the current Portfile. In that case, you can use "-o /path/to/Portfile" to "help" upt-macports. To summarize, it is usually ok to run:

    $ upt package -f pypi -b macports -u requests

But sometimes, you'll need

    $ upt package -f pypi -b macports -u -o ./python/py-something/Portfile aweirdpackage

Ideally we could find all packages for which the "PyPI name to Macports name" conversion fails and fix them, thus making sure we never have to use -o again.

> In other words if you specify -o ~/my-macports-clone it will generate the Portfile in ~/my-macports-clone/python/py-<packagename>/Portfile. It behaves quite differently here in the sense that you'd need to supply a full path including the Portfile part.

Oh, good point.

My initial idea was that "--output" could be used the same way for creating/updating packages across all upt backends, but somehow it became a bit of a mess. We could make sure that:

    $ upt package -f pypi -b macports -u -o ./python/py-something/Portfile aweirdpackage

and 

    $ upt package -f pypi -b macports -u -o ./python/py-something aweirdpackage

both work, but it would still not be optimal :/

PR #71, opened 2021-09-11
Do not crash when no archives could be found.

### LINKED ISSUE #70: Can't create a template for `pyzbar`
Installed:
```
  py39-upt-cpan @0.6.1_0 (active)
  py39-upt-macports @0.1-20200607_0 (active)
  py39-upt-pypi @0.6_0 (active)
  py39-upt-rubygems @0.4.1_0 (active)
```
when run `upt package -f pypi -b macports -o . -r pyzbar` it crashed:
```
[INFO    ] [Backend] Creating MacPorts package for pyzbar
[INFO    ] [Backend] Found license MIT
[ERROR   ] [Backend] Could not determine the type of the source archive
Traceback (most recent call last):
  File "/opt/local/bin/upt", line 33, in <module>
    sys.exit(load_entry_point('upt==0.12', 'console_scripts', 'upt')())
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/upt/upt.py", line 497, in main
    package(name, version, frontend, backend, args.output,
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/upt/upt.py", line 382, in package
    backend.create_package(upt_pkg, output=output)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/upt_macports/upt_macports.py", line 258, in create_package
    packager.create_package(upt_pkg, output)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/upt_macports/upt_macports.py", line 20, in create_package
    portfile_content = self._render_makefile_template()
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/upt_macports/upt_macports.py", line 57, in _render_makefile_template
    return template.render(pkg=self)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/jinja2/environment.py", line 1304, in render
    self.environment.handle_exception()
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/jinja2/environment.py", line 925, in handle_exception
    raise rewrite_traceback_stack(source=source)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/upt_macports/templates/python.Portfile", line 1, in top-level template code
    {% extends 'base.Portfile' %}
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/upt_macports/templates/base.Portfile", line 41, in top-level template code
    checksums           rmd160  {{ pkg.upt_pkg.archives[0].rmd160 }} \
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/jinja2/environment.py", line 474, in getattr
    return getattr(obj, attribute)
jinja2.exceptions.UndefinedError: list object has no element 0
```

Steap:
Close: #70

### DIFF
--- a/upt_macports/templates/base.Portfile
+++ b/upt_macports/templates/base.Portfile
@@ -38,9 +38,14 @@ long_description    ${description}
 
 {% block dist_info %}
 {% endblock %}
+{% if pkg.upt_pkg.archives %}
 checksums           rmd160  {{ pkg.upt_pkg.archives[0].rmd160 }} \
                     sha256  {{ pkg.upt_pkg.archives[0].sha256 }} \
                     size    {{ pkg.upt_pkg.archives[0].size }}
+{% else %}
+# TODO: The archive was not found (format: {{ pkg.archive_format }}).
+#       The checksums could therefore not be computed.
+{% endif %}
 {% block versions %}
 {% endblock -%}
 


reneeotten:
@catap that already happens now with some logging, but perhaps we should update the message a bit:
```
[ERROR   ] [Backend] Could not determine the type of the source archive
```

@Steap thanks for the fix here and the PR with the update function! Instead of changing things in the template so that it doesn't crash in such an unfriendly way; I would probably prefer to just to `exit` when the archive couldn't be found - there is nothing we can do with `upt` or in MacPorts if that is the case. 

So how about doing this in `upt_macports.py` instead:
```
            except upt.ArchiveUnavailable:
                self.logger.error('Could not determine the type of the source archive') # noqa
-               return 'unknown'
+               sys.exit(1)
```

catap:
@reneeotten I dissagree with idea to `exit`. Let me explain. Right now `utp` just created a port without checksum and when it works with `-r` it continue to create dependencies ports. Am I wrong?

Steap:
@reneeotten As @catap pointed out, when using "-r" for recursive packaging, it might be interesting not to exit.

upt can do a few things when the tarball cannot be found: for instance, the dependencies are computed from the wheel, and some of the metadata are computed from a JSON file served by PyPI. This does mean that the packager will need to find the tarball somewhere else (possibly on github, for instance).

It should be very rare that tarballs are not uploaded to PyPI, though.

PR #72, opened 2022-05-09
Update to python 3.10, remove livecheck none

mojca:
See https://github.com/macports/macports-ports/pull/14793#pullrequestreview-966363348


reneeotten:
thanks @mojca - looks good. I had these (and a few more changes) lined up, but got busy with family-life ;) I still need to do a bit more testing of #69 - Cyril poked me again about that.... I *think* we should just merge that PR, but if you have a few cycles to do some testing and provide feedback that would be great!

PR #74, opened 2023-09-09
Add `platforms {darwin any}` to base.Portfile

catap:


reneeotten (commented):
that doesn't seem to be the correct place; this line is (more) specific to Python ports because of the Framework installation location. For most other ports it should just be "any" as dar as I can remember; but checking Josh his e-mail on the usage when this was introduced is probably a good idea.

jmroot:
Perl also uses a platform-specific directory structure. Another complication is that being noarch doesn't guarantee that a port doesn't install different files on different OS versions.

For Python, I think it should be possible to parse some metadata or at least the filenames of the available wheels in order to figure out if the package is architecture and/or platform independent.

PR #76, opened 2025-08-11
Update supported versions of Python

Steap:


pmetzger:
Failure seems peculiar.

Steap:
Let us merge #77 instead

